PR #41, opened 2023-03-24
Basic support for `@constructor`

antroseco:
We still need `this = { ... }` support.
If any constructors are defined for the struct you are trying to initialize, you can no longer use the initializer list syntax.


DaveDuck321:
This is VERY out-of-date :cry: tried to cherry-pick but couldn't even find the correct lines.

PR #88, opened 2023-07-16
Remove the annoying syntax

DaveDuck321:
We no longer require semicolons after every top-level statement and `if`/ `for`/ `while` no longer require parens around the condition.

I'm a bit unsure about the aesthetic of foreign functions and typedef aliases... but it's unambiguous so let's give it a shot.

### DIFF
--- a/std/syscalls.c3
+++ b/std/syscalls.c3
@@ -1,56 +1,57 @@
-@require_once "util.c3";
-@require_once "sys/uio.c3";
+@require_once "util.c3"
+@require_once "sys/uio.c3"
 
-foreign _syscall_0 : (syscall_number: i64) -> i64;
-foreign _syscall_1 : (a: i64, syscall_number: i64) -> i64;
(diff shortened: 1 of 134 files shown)


std/syscalls.c3
```diff
@@ -1,56 +1,57 @@
...
-@require_once "sys/uio.c3";
+@require_once "util.c3"
+@require_once "sys/uio.c3"
 
-foreign _syscall_0 : (syscall_number: i64) -> i64;
-foreign _syscall_1 : (a: i64, syscall_number: i64) -> i64;
-foreign _syscall_2 : (a: i64, b: i64, syscall_number: i64) -> i64;
-foreign _syscall_3 : (a: i64, b: i64, c: i64, syscall_number: i64) -> i64;
-foreign _syscall_6 : (a: i64, b: i64, c: i64, d: i64, e: i64, f: i64, syscall_number: i64) -> i64;
+
```
antroseco, line 4: Extra newline

antroseco (changes requested):

antroseco (approved):

PR #94, opened 2023-08-06
Self-hosted parser

DaveDuck321:


DaveDuck321:
This isn't even close to ready but some feedback would be useful.

I've resisted adding language features but the consequences is some janky type erasure, hacky global allocator (see the Cursor type), and non-ergonomic chars.

std/string.c3
```diff
@@ -19,3 +23,42 @@ function [@Len] make_cstring : (literal : u8[@Len]&) -> CString = {
...
+        return char | __builtin_narrow<u8>(0x20);
+    }
+    return char;
+}
+
+function ascii_to_upper : (char : u8) -> u8 = {
+    // TODO: support &&
+    // TODO: support char literals
+    if (as_arithmetic(char) >= as_arithmetic("a"[0])) | (as_arithmetic(char) <= as_arithmetic("z"[0])) {
+        return char & __builtin_narrow<u8>(0x20);
```
DaveDuck321, line 61: This is a bug. I believe it should be `char & ~0x20`

parser.c3
```diff
@@ -0,0 +1,658 @@
...
+
+    let allocator : Allocator = initialize_allocator();
+    let file_contents : Vector<u8> = make_vector<u8>(&allocator);
+
+    // 1) Read the entire file
+    {
+        let read_buffer : u8[1024];
+
+        // TODO: `do while` to avoid temp value? Or `break` to make this prettier
+        let read_data : StringView = make_span("uninitialized");
```
antroseco, line 646: something like `"text"sv` would be nice eventually

parser.c3
```diff
@@ -0,0 +1,658 @@
...
+    }
+
+    let args : Vector<Argument> = cursor:parse_comma_separated_named_members(make_span(")"));
+
+    if !cursor:try_skip_next_and_surrounding_white_space("->") {
+        syntax_error(cursor.position, "TODO: expected '->'");
+    }
+
+    let return_type : Type = cursor:parse_type();
+    let scope : Scope = cursor:parse_scope();
```
antroseco, line 598: Did you drop the `=`?

parser.c3
```diff
@@ -0,0 +1,658 @@
...
+    }
+    return generics;
+}
+
+function parse_type : (cursor : Cursor&, context : LegalEndOfToken) -> Type = {
+
+}
+
+// BUG: Both `cursor` and `Cursor` are highlighed on the next line
+function parse_typedef : (cursor : Cursor&) -> Typedef = {
```
antroseco, line 465: ah lol, it's cause `parse_typedef : (cursor` looks like a typedef

parser.c3
```diff
@@ -0,0 +1,658 @@
...
+    }
+
+    // Nothing to skip
+    return false;
+}
+
+function try_skip_white_space : (cursor : Cursor&) -> bool = {
+    if ascii_is_white_space(cursor:get_char()) {
+        cursor.position += 1;
+        cursor:try_skip_white_space();
```
antroseco, line 260: Can we do this without O(n) space pls

parser.c3
```diff
@@ -0,0 +1,658 @@
...
+    }
+    return true;
+}
+
+function [T] try_skip_next_and_surrounding_white_space : (cursor : Cursor&, string : T) -> bool = {
+    cursor:try_skip_white_space();
+    let result : bool = cursor:try_skip_next();
+    cursor:try_skip_white_space();
+    return result;
+}
```
antroseco, line 210: arg `string` is unused

antroseco (commented):
Looks good so far!

std/io.c3
```diff
@@ -29,3 +32,28 @@ function[@Len] puts : (str: u8[@Len]&) -> int = {
     // TODO: and errno shall be set to indicate the error.
     return EOF();
 }
+
+function[@Len] puts : (str: u8[@Len]&) -> int = {
+    return puts({&str, @Len});
+}
```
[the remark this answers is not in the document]
antroseco: nvm, this is what breaks CI :(
guess we need a smarter compiler

parser.c3
DaveDuck321: :upside_down_face: maybe smart. I'll do the rest first

parser.c3
DaveDuck321: Fair enough. I'll remove the comments that I've already opened issues for

parser.c3
DaveDuck321: Not intentionally :stuck_out_tongue_winking_eye:

parser.c3
DaveDuck321: agreed!

parser.c3
```diff
@@ -0,0 +1,658 @@
...
+
+function parse_next_top_level : (cursor : Cursor&) -> TopLevelToken = {
+    // typedef 'BUG IN SYNAX HIGHLIGHTING... try removing this quoted text'
+    if cursor:is_next("typedef") {
+        // BUG IN SYNTAX HIGHLIGHTING: remove this line
+        let typedef : Typedef& = cursor.allocator:allocate<Typedef>();
+
+        // TODO: this is a copy from stack to heap... is it obvious enough?
+        typedef = cursor:parse_typedef(); // TODO: 'typedef' is highlighted
```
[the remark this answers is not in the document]
DaveDuck321: Hmm, no technical reason to block it. Atm only `while`/`for`/`if`/`return`/... are going to be invalid.

But also... its ugly so I'll block it (also `function` too I suppose)

std/io.c3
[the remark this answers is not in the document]
DaveDuck321: Lol, I added the explicit specialization and now we're somehow breaking codegen.

std/syscalls.c3
```diff
@@ -1,54 +1,43 @@
...
-    return 0x20;
-}
-
 function sys_mmap : (addr: isize, length : isize, prot: u32, flags: u32, fd: i32, offset: isize) -> iptr = {
     return _syscall_6(addr, length, as_arithmetic(prot), as_arithmetic(flags), fd, offset, 9);
 }
 
-function sys_munmap : (addr: isize, length : isize) -> void = {
-    _syscall_2(addr, length, 11);
+function sys_munmap : (addr: isize, length : isize) -> isize = {
```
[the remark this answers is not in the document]
DaveDuck321: Great, I'll make it consistent I suppose.

std/string.c3
```diff
@@ -0,0 +1,21 @@
...
+    const null : u8[1]& = "\0"; // TODO: add char
+
+    while str.data[len] != null[0] {
+        len = len + 1;
+    }
+
+    return len;
+}
+
+// TODO: this is kinda ugly, we need partial functions
```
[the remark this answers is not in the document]
DaveDuck321: Hmm, took me a second to understand. This is a bad comment imo.

I think I wanted something like: `make<CString>()` rather than `make_cstring()`. But to do that we'd need to allow functions to have partial specializations (so the other generic arguments could be inferred).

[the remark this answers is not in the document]
antroseco: Ah, right. I remember you talking about that.

parser.c3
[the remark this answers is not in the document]
antroseco: for purely aesthetic reasons

tests/regressions/pattern_match_heap_array.c3
```diff
@@ -9,4 +9,6 @@ function main : () -> int = {
 /// @COMPILE; EXPECT ERR
 /// File '*.c3', line 5, in 'function main: () -> int'
 /// Error: overload resolution for function call 'puts(u8[[]&[]])' failed
-/// No overloads available
+/// Available overloads:
+/// - function puts: (Span<u8>) -> int
+/// - function puts: (CString) -> int
```
antroseco, line 14: Wait, this looks like a pretty big change.  Are we now offering to initialize a struct implicitly, without wrapping the argument in `{ }`?

DaveDuck321: No, we still require manual initialization. These are just the new (non-templated) overloads specified in `io.c3`.

But also this is probably a bug tbh, we should also list the generic functions in the list of available overloads.

parser.c3
```diff
@@ -0,0 +1,658 @@
...
+        return {2, ref_to_addr(&typedef)};
```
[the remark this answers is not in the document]
DaveDuck321: The kinda implicit heap copy or calling a variable `typedef`?

parser.c3
```diff
@@ -0,0 +1,658 @@
...
+    return result;
+}
+
+function parse_graphene_function : (cursor : Cursor&) -> FunctionDefinition = {
+    runtime_assert(cursor:try_skip_next_and_surrounding_white_space("function"));
+
+    // Generic definitions
+    let generics : Vector<GenericDefinition>; // Uninitialized
+
+    if cursor:has_next("[") {
```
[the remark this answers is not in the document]
DaveDuck321: Lol, ur right, its `is_next`, I was misremembering when I typed it.

I haven't tried to compile yet :upside_down_face:, there's a fair few functions without return.

tests/regressions/pattern_match_heap_array.c3
antroseco: Oh right, duh. But you're right, we should be listing the generic function too.

parser.c3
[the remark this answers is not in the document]
antroseco: Calling a variable `typedef`

parser.c3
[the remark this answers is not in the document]
antroseco: Ah, let's see how many bugs you have to fix before it compiles then!

parser.c3
DaveDuck321: What do you think of `sv("text")` instead?

It's slightly more verbose but there's no magic because unlike c++ we can actually already pattern match on strings, no need for the extra language feature.

parser.c3
DaveDuck321: Done

std/string.c3
```diff
@@ -1,10 +1,29 @@
 @require_once "arithmetic.c3"
 @require_once "assert.c3"
+@require_once "iterators.c3"
 @require_once "util.c3"
 @require_once "wrappers.c3"
 
 typedef StringView : Span<u8>
 
+@operator == : (lhs : StringView, rhs : StringView) -> bool = {
```
antroseco, line 13: Wasn't `StringView` a `Span<u8>`? Can't we then generalize this for `Span<T>`?

parser/parser.c3
```diff
@@ -913,11 +914,51 @@ function parse_function_call : (cursor : Cursor&, prefix : Expression) -> Functi
...
+
     if !cursor:is_next("<") {
         return false;
     }
-    return true;
+
+    let brackets_stack : Vector<u8> = make_vector<u8>(&cursor.allocator);
+    let lookahead : Cursor = cursor;
+    while !lookahead:is_next("\0") {
+        if is_one_of(lookahead:get_char(), "<"[0], "("[0], "{"[0], "["[0]) {
```
antroseco, line 932: Curious to see what this function call optimizes to

antroseco (approved):
Very nice!

parser/parser.c3
DaveDuck321: I'll check... But it's full of dynamic allocation so not great (I think it's opaque to llvm until we annotate it). I could defo make it faster by removing the allocation and making it recursive, but unless it's a bottleneck I don't think it's worth the risk of a stack overflow.

DaveDuck321: I'm curious about `is_one_of` too, I hope it's equivalent to whatever c produces with a switch block.

std/string.c3
DaveDuck321: Good point. That's seems useful. We don't have the SFINAE constructs to limit it to spans with an elementwise `==` but it'll be fine for now.

parser/file_info.c3
```diff
@@ -24,18 +24,25 @@ function generate_position_lookup_table : (program : StringView, allocator: Allo
...
     return table.offsets.length - 1;
 }
 
 function lookup_position : (table: PositionLookupTable&, offset : isize) -> FilePosition = {
-    let line_number : isize = table:lookup_line_number(offset);
-    let column_number : isize = offset - table.offsets:get(line_number);
-    return {line_number, column_number, offset};
+    let line_number : isize = table:lookup_zero_indexed_line_number(offset);
+
+    // TODO: I'm using this pattern a lot, this should be a single statement
```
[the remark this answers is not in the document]
DaveDuck321: Ternary syntax is a bit clunky imo, you can't have multiple lines of code.

It'd kinda be nice if we could `yield ` from an `if` statement/ the other control blocks.

parser/parser.c3
antroseco: I've checked `is_one_of` with a bunch of `int`s... If they are all known at compile time, then LLVM sees right through it and computes the answer for you. The recursive function calls are all inlined into the top-level call to `is_one_of`, resulting in a series of `cmp/je` instructions.

std/string.c3
antroseco: Maybe this should count as a "substitution failure" for the purposes of SFINAE

parser/file_info.c3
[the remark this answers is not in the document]
antroseco: `yield` is cool

antroseco:
Any chance we can break this up into smaller PRs and start merging them individually?

DaveDuck321:
> Any chance we can break this up into smaller PRs and start merging them individually?

Yeah, I'll try... gotta keep the history clean. The scope kinda took me by surprise.

DaveDuck321:
Closed in favor of #101, #102, and #103

Issue #99, opened 2023-08-16
`sys_m{map, unmap}` have incorrect signatures


antroseco:
https://pubs.opengroup.org/onlinepubs/009604499/functions/mmap.html
https://pubs.opengroup.org/onlinepubs/000095399/functions/munmap.html

First argument should be `iptr`, not `isize`.
`munmap` should also have a return type.

idiotcommerce:
hi i want to work on this.  but small doubt. 

First argument should be iptr not isize.

don't take me wrong. i will show some images of code please help on that.

these are the functions you used as first argument as   ISIZE.  tell me where should i change.  not on this images 
very very sorry for that.

1  

![image](https://github.com/DaveDuck321/GrapheneLang/assets/121589808/d72216c7-e2e7-416a-8c7a-13d54fb7ee02)

2  

![image](https://github.com/DaveDuck321/GrapheneLang/assets/121589808/c825118a-d3b6-4cfe-a9bf-8d5045dfc837)

3

![image](https://github.com/DaveDuck321/GrapheneLang/assets/121589808/08ed8925-ce48-4198-98de-197042d09c25)

antroseco:
Hi, and welcome to the project! I had already fixed this issue in one of the open PRs (https://github.com/DaveDuck321/GrapheneLang/pull/100/commits/20621de452b5d51454dc8784644f6f2f1214f7c0), but feel free to work on something else.

idiotcommerce:
Ok ok

PR #100, opened 2023-08-16
Add ARM support

antroseco:
64-bit support is almost ready. 32-bit ARM is a bit harder to work on as we can't really run our tests yet.

There's still a bit of work to do on the runtimes.


std/aarch64_linux/syscalls.c3
```diff
@@ -0,0 +1,32 @@
...
+    return _syscall_3(fd, iov_addr, iovcnt, /* writev */ 66);
+}
+
+function sys_mmap : (addr: isize, length : isize, prot: u32, flags: u32, fd: i32, offset: isize) -> iptr = {
+    return _syscall_6(addr, length, as_arithmetic(prot), as_arithmetic(flags), fd, offset, /* mmap */ 222);
+}
+
+function sys_munmap : (addr: isize, length : isize) -> void = {
+    _syscall_2(addr, length, /* munmap */ 215);
+}
```
DaveDuck321, line 32: Can we fix #99 with this PR?

std/arm_linux/runtime.S
```diff
@@ -0,0 +1,63 @@
...
+
+.global _syscall_1
+_syscall_1:
+    # Syscall number is passed via r7, but r7 must be preserved, so save it on
+    # the stack.
+    push    {r7}
+    mov     r7, r1
+    swi     0
+    pop     {r7}
+    bx      lr
```
DaveDuck321, line 53: Are r4-r7 caller saved by arm/EABI? If not they should also be saved here

std/sys/mman.c3
```diff
@@ -0,0 +1,15 @@
...
+function SYS_PROT_READ  : () -> u32 = { return 0x1; } // Page can be read.
+function SYS_PROT_WRITE : () -> u32 = { return 0x2; } // Page can be written.
+function SYS_PROT_EXEC  : () -> u32 = { return 0x4; } // Page can be executed.
+function SYS_PROT_NONE  : () -> u32 = { return 0x0; } // Page can not be accessed.
+
+function SYS_MAP_SHARED          : () -> u32 = { return 0x1; } // Share changes.
+function SYS_MAP_PRIVATE         : () -> u32 = { return 0x2; } // Changes are private.
+function SYS_MAP_SHARED_VALIDATE : () -> u32 = { return 0x3; } // Share changes and validate extension flags.
+
+function SYS_MAP_ANONYMOUS : () -> u32 = { return 0x20; } // Don't use a file.
```
DaveDuck321, line 15: This is much cleaner, nice

tests/arrays/alignment_aarch64_linux.c3
```diff
@@ -0,0 +1,14 @@
...
+    // Arrays use the alignment of their elements.
+    let small : int[3];
+
+    // Irrespective of the total array size.
+    let large : int[4];
+
+    return 0;
+}
+
+/// @FOR aarch64_linux
```
DaveDuck321, line 11: Ooo, smart. Depending on how many of these we want, a target directory might be nice. This is good for now tho.

tests/misc/fork.c3
```diff
@@ -14,6 +14,7 @@ function main : () -> int = {
     return 0;
 }
 
+/// @FOR x86_64_linux
```
DaveDuck321, line 17: Is this just temporary until we've implemented a `clone3` based `fork` for arm?

DaveDuck321 (approved):
Very cool

std/arm_linux/runtime.S
antroseco: r4-r11 and sp are callee-saved, but we are only touching r7, right?

tests/misc/fork.c3
antroseco: Yes, although that's not really a priority

tests/arrays/alignment_aarch64_linux.c3
antroseco: Yeah, I know. I just didn't feel like replicating all the subdirectories etc multiple times. If we have too many of them we'll make separate directories.

std/aarch64_linux/syscalls.c3
antroseco: Done

std/arm_linux/runtime.S
DaveDuck321: Ok, that makes sense, no need then. I thought the kernel might not bother saving them given that it already has a different calling convention.

std/sys/mman.c3
```diff
@@ -0,0 +1,35 @@
...
+    addr: iptr,
+    length: isize,
+    prot: u32,
+    flags: u32,
+    fd: int,
+    offset: off_t
+) -> iptr = {
+    return sys_mmap(addr, length, as_arithmetic(prot), as_arithmetic(flags),
+                    fd, offset);
+}
```
DaveDuck321, line 31: Is this purely so the interface of `sys_mmap` matches the Linux calling convention or does it also help us with platform compatibility?

DaveDuck321 (approved):

std/sys/mman.c3
```diff
@@ -0,0 +1,35 @@
...
+// TODO: support enums (or similar)
+
+function SYS_PROT_READ  : () -> u32 = { return 0x1; } // Page can be read.
+function SYS_PROT_WRITE : () -> u32 = { return 0x2; } // Page can be written.
+function SYS_PROT_EXEC  : () -> u32 = { return 0x4; } // Page can be executed.
+function SYS_PROT_NONE  : () -> u32 = { return 0x0; } // Page can not be accessed.
+
+function SYS_MAP_SHARED          : () -> u32 = { return 0x1; } // Share changes.
+function SYS_MAP_PRIVATE         : () -> u32 = { return 0x2; } // Changes are private.
+function SYS_MAP_SHARED_VALIDATE : () -> u32 = { return 0x3; } // Share changes and validate extension flags.
```
DaveDuck321, line 17: This is going to conflict with a similar change I made in #101. I like this one more tho, lets keep this

std/arm_linux/runtime.S
antroseco: To be honest, not really sure what the kernel is supposed to do. Linux claims that only `lr` is clobbererd (https://github.com/torvalds/linux/blob/93f5de5f648d2b1ce3540a4ac71756d4a852dc23/tools/include/nolibc/arch-arm.h#L57) and indeed, the syscall handler starts by pushing `r1-r12` to the stack. On the other hand, glibc saves `r4-r7` only (not `lr`) (https://github.com/bminor/glibc/blob/4290aed05135ae4c0272006442d147f2155e70d7/sysdeps/unix/sysv/linux/arm/syscall.S#L26). So... I think I won't save any other registers... as both implementations seem to work just fine... If this turns out to be a problem, then it's an easy fix.

PR #101, opened 2023-08-18
Standard library support required for the parser

DaveDuck321:


std/format.c3
```diff
@@ -0,0 +1,145 @@
...
+
+function [IO, NextArg] writef : (io : IO&, format_string : StringView, next : NextArg) -> void = {
+    let next_format_string : StringView = write_next_format_pair(&io, format_string, next);
+    writef(&io, next_format_string);
+}
+
+function [IO, NextArg, Args...] writef : (io : IO&, format_string : StringView, next_arg: NextArg, args : Args...) -> void = {
+    let next_format_string : StringView = write_next_format_pair(&io, format_string, next_arg);
+
+    // UFCS breaks with parameter packs!!
```
antroseco, line 128: Huh, that's a bug then

std/format.c3
```diff
@@ -0,0 +1,145 @@
...
+
+    let result : StringView = {&buffer.data, length};
+
+    // XXX: its a bit unfair to call it a `StringView` when its mutated in-place
+    result:reverse();
+    io:consume_length(length);
+    return result;
+}
+
+function [IO] write : (io : IO&, number : u32) -> TypeIf<StringView, IsStreamableTextIO<IO>> = {
```
antroseco, line 52: Did you change your mind on "stringify"? I thought the flexibility was nice.

std/math.c3
```diff
@@ -0,0 +1,9 @@
+@require_once "std/type_traits.c3"
+
+function [T] abs : (value : T) -> TypeIf<T, IsIntegral<T>> = {
+    if value < 0 {
+        return -value;
+    } else {
+        return value;
+    }
```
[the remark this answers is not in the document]
DaveDuck321: That was a good read, I've had fun playing around with it :smiley:. I do think is a premature optimization tho.

1) LLVM at -O1 is correctly able to deduce both algorithms are an `abs` (compiling them to the same IR call). The only difference is that there are MORE debug symbols in the branchless version since more variables are collapsed (it will be slower if anything).
image


2) The AMD64 output actually more closely resembles the version WITH branches: LLVM is able to outsmart us, see what we're trying to do with the bit-twiddling, and remove our optimization.
image

Let's just keep the branches for readability, LLVM can see it's an `abs` so it'll codegen the optimal assembly for each target regardless.

std/format.c3
DaveDuck321: K, opened in #105

std/format.c3
DaveDuck321: Yeah, I thought `stringify` was nice too, the problem is it becomes kinda unwieldy unless we want to let it do dynamic allocations.

Let's say we want to implement something like `sprintf`, how do we know what size of buffer to give it? Is it always the caller's responsibility? That seems like a c-style memory safety issue.

Let's say we allow dynamic allocation (so the `stringify` can choose its own buffer), now file/ tty io will become kinda ugly because instead of writing blocks we've started allocating huge buffers and writing the entire thing at once (also now we've got to pass an allocator to each stringify call).

Using streamables instead doesn't require any dynamic allocation for file-backed IO but without any code duplication we can also use it to write a dynamically allocating stringify/ StringStream:

```rust
let io_global_allocator : ...;

function request_buffer : (dynamic_io : StringStream&, buffer_length : isize) -> Span<u8> = {
	dynamic_io.underlying_vector:grow_capacity(buffer_length);
	return dynamic_io.underlying_vector:data();
}

function stringify : (a : MyType) -> StringView = {
	let io : StringStream = ...;
}
```

std/wrappers.c3
```diff
@@ -38,3 +38,36 @@ typedef [T] Span : {
     data : T[&],
     length : isize,
 }
+
+typedef [T] SpanIter : {
+    span : Span<T>,
+    index : isize,
+}
+
+function [T] slice_from_offset : (span : Span<T>&, start_offset : iptr) -> Span<T> = {
```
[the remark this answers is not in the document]
DaveDuck321: Whoops, sorry about that :cry:

graphene_parser.py
```diff
@@ -557,6 +558,9 @@ def try_deduce_specialization(
             # Convert the deduced mapping into a specialization
             deduced_specialization: list[cg.SpecializationItem] = []
             for generic in generic_definitions:
+                if generic.name not in deduced_mapping:
+                    raise PatternMatchDeductionFailure(fn_name, generic.name)
```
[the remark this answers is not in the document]
DaveDuck321: Fixed in f54af9d92cfb9f8bf3c748db43e7df1d63612191

std/math.c3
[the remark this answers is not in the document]
antroseco: Hah, fair enough. Kinda amazed that it can see straight through the bit-hacks!

Interesting to note that for riscv, the assembly output is essentially the "patented version" from that website

std/format.c3
antroseco: > Let's say we allow dynamic allocation (so the stringify can choose its own buffer), now file/ tty io will become kinda ugly because instead of writing blocks we've started allocating huge buffers and writing the entire thing at once (also now we've got to pass an allocator to each stringify call).

Hmm... Can't we have `stringify` return a `std::string` equivalent (kinda hard without destructors but let's say we have RAII one day)
Then you can do scatter/gather I/O with `writev` and it's not even that bad

PR #102, opened 2023-08-18
Conform to a new, generic parser interface

DaveDuck321:
This will allow the self-hosted parser to be easily swapped in: all it needs to do is conform to the JSON interface (currently based closely on the lark tree).

In the spirit of removing the lark dependency, I've also reimplemented the standard lark algorithms we rely on (eg. `Transformer` and `Interface`). These are a bit awkward and in the future I think it'd be worth investing time into making everything more ergonomic (which would also improve the type hinting).


graphene_parser.py
```diff
@@ -950,28 +955,30 @@ def _split_function_call_args(
...
-        return self._function_call_impl(fn_name, specialization_tree, args)
+        return self._function_call_impl(fn_name.value, specialization_tree, args)
 
-    @v_args(inline=True)
+    @inline_arguments
     def ufcs_call(
         self, this_tree: Tree, name_tree: Tree, *all_args: Any
     ) -> FlattenedExpression:
         # TODO support unpacking.
+        # TODO `pack_name` is unused
```
antroseco, line 981: Looks like you found that bug!

parser/bootstrap/lark_parser.py
```diff
@@ -0,0 +1,101 @@
...
+        "name": token.type,
+        "value": token.value,
+        "meta": {
+            "start": {"line": token.line, "column": token.column},
+            "end": {"line": token.end_line, "column": token.end_column},
+        },
+    }
+
+
+def tree_to_dict(tree: Tree) -> dict[str, Any]:
```
antroseco, line 30: This might be cleaner with a `json.JSONEncoder`

parser/lexer_parser.py
```diff
@@ -0,0 +1,114 @@
...
+class Interpreter:
+    def __init__(self, visit_tokens=False) -> None:
+        self._visit_tokens = visit_tokens
+
+    def visit(self, tree: Tree | Token):
+        if not hasattr(self, tree.name):
+            if isinstance(tree, Token):
+                return tree
+
+            return self.visit_children(tree)
```
antroseco, line 80: This deviates slightly from lark's interface

> Unlike Transformer and Visitor, the Interpreter doesn’t automatically visit its sub-branches. The user has to explicitly call visit, visit_children, or use the @visit_children_decor. This allows the user to implement branching and loops.

It's fine if this is intentional

antroseco (changes requested):
Looks good, but can you manually rebase on `main` please? This is only one commit from what I can see, but GitHub still shows the previous 15.

graphene_parser.py
DaveDuck321: Lol, I didn't read the TODO there

parser/lexer_parser.py
DaveDuck321: Lark actually does the same thing here (despite the docs)... If you run the interpreter and Lark parses a tree that doesn't correspond to a method of your specialized Interpreter, it just visits the children instead (see `__default__()` in the code below)...

Our current `graphene_parser.py` actually requires this feature since we don't actually bother matching the `program` grammar, instead, we're only interested in its top-level children (which lark automatically iterates through).

```python
class Interpreter(_Decoratable, ABC, Generic[_Leaf_T, _Return_T]):
    """Interpreter walks the tree starting at the root.

    Visits the tree, starting with the root and finally the leaves (top-down)

    For each tree node, it calls its methods (provided by user via inheritance) according to ``tree.data``.

    Unlike ``Transformer`` and ``Visitor``, the Interpreter doesn't automatically visit its sub-branches.
    The user has to explicitly call ``visit``, ``visit_children``, or use the ``@visit_children_decor``.
    This allows the user to implement branching and loops.
    """

    def visit(self, tree: Tree[_Leaf_T]) -> _Return_T:
        # There are no guarantees on the type of the value produced by calling a user func for a
        # child will produce. So only annotate the public method and use an internal method when
        # visiting child trees.
        return self._visit_tree(tree)

    def _visit_tree(self, tree: Tree[_Leaf_T]):
        f = getattr(self, tree.data)
        wrapper = getattr(f, 'visit_wrapper', None)
        if wrapper is not None:
            return f.visit_wrapper(f, tree.data, tree.children, tree.meta)
        else:
            return f(tree)

    def visit_children(self, tree: Tree[_Leaf_T]) -> List:
        return [self._visit_tree(child) if isinstance(child, Tree) else child
                for child in tree.children]

    def __getattr__(self, name):
        return self.__default__

    def __default__(self, tree):
        return self.visit_children(tree)
```

parser/bootstrap/lark_parser.py
DaveDuck321: I looked into this a bit... Looks like there's no way to do it more concisely than this without a more strongly typed interface. I plan to make this better once we've removed lark tho.

parser/lexer_parser.py
antroseco: Ah, I see. yeah that makes sense

PR #103, opened 2023-08-18
Self-host the parser

### LINKED ISSUE #10: Remove Lark dependency
Selfhosting is kinda cool but I doubt it would be fun rewriting the whole compiler. How about we aim to selfhost the *lark* lexer-parser (and add python bindings)?

This will also be a good tool to benchmark the codegen. There are a few hints I want to add but can't justify without verifying that they improve the output.

DaveDuck321:
Closes #10 and improves performance enough to make the compiler fun to iterate with.

### DIFF
--- a/codegen/interfaces.py
+++ b/codegen/interfaces.py
@@ -275,6 +275,22 @@ SpecializationItem = Type | int
 GenericMapping = dict[str, SpecializationItem]
 
 
+def do_specializations_match(
+    s1: list[SpecializationItem], s2: list[SpecializationItem]
+) -> bool:
+    if len(s1) != len(s2):
+        return False
+
+    for item1, item2 in zip(s1, s2):
+        if isinstance(item1, Type) != isinstance(item2, Type):
+            return False
+
+        if item1 != item2:
+            return False
+
+    return True
+
+
 def format_specialization(specialization: list[SpecializationItem]) -> str:
     if len(specialization) == 0:
         return ""
--- a/parser/parser.c3
+++ b/parser/parser.c3
@@ -228,7 +228,7 @@ function get_next_hex_token : (cursor : Cursor&) -> ProgramSlice = {
             token.end_offset += 1;
         } else {
             // Otherwise its a syntax error
-            syntax_error(&cursor, sv("illegal digit '%' in hex constant"), decode_rune(next_char));
+            syntax_error(&cursor, sv("illegal digit '%' in hex constant"), decode_utf8(next_char));
         }
     }
     if token.end_offset - token.start_offset == 2 {
@@ -266,8 +266,7 @@ function get_next_decimal_token : (cursor : Cursor&) -> ProgramSlice = {
             token.end_offset += 1;
         } else {
             // Otherwise its a syntax error
-            let char : Rune = {next_char};
-            syntax_error(&cursor, sv("illegal digit '%' in decimal constant"), char);
+            syntax_error(&cursor, sv("illegal digit '%' in decimal constant"), decode_utf8(next_char));
         }
     }
     if token.end_offset - token.start_offset == 0 {
@@ -343,7 +342,7 @@ function get_next_operator : (cursor : Cursor&) -> ProgramSlice = {
         return {start_position, start_position + 1};
     }
 
-    syntax_error(&cursor, sv("illegal operator '%'"), decode_rune(cursor:get_char()));
+    syntax_error(&cursor, sv("illegal operator '%'"), decode_utf8(cursor:get_char()));
     return {0, 0};
 }
 
@@ -587,7 +586,7 @@ function parse_type_prefix : (cursor : Cursor&) -> Type = {
         return {5, ref_to_addr(&struct_type), meta(start_offset, struct_type.meta.end_offset)};
     }
 
-    syntax_error(&cursor, sv("unexpected symbol '%', this cannot be used at the start of a type"), decode_rune(cursor:get_char()));
+    syntax_error(&cursor, sv("unexpected symbol '%', this cannot be used at the start of a type"), decode_utf8(cursor:get_char()));
 
     // Unreachable
     // TODO: maybe allow `undefined` return
@@ -927,7 +926,7 @@ function parse_non_operator_expression_prefix : (cursor : Cursor&) -> Expression
         return {4, ref_to_addr(&variable_access), get_meta(name)};
     }
 
-    syntax_error(&cursor, sv("unexpected symbol '%', cannot start an expression"), decode_rune(cursor:get_char()));
+    syntax_error(&cursor, sv("unexpected symbol '%', cannot start an expression"), decode_utf8(cursor:get_char()));
     return {0, 0, meta(0, 0)};
 }
 
--- a/parser/parser.c3
+++ b/parser/parser.c3
@@ -23,10 +23,6 @@ function slice : (cursor : Cursor&, length : isize) -> ProgramSlice = {
     return {cursor.position, cursor.position + length};
 }
 
-function slice : (cursor : Cursor&, start : isize, length : isize) -> ProgramSlice = {
-    return {start, start + length};
-}
-
 function ascii_is_text_token_start_char : (char : u8) -> bool = {
     return ascii_is_alpha(char) | (char == "_"[0]);
 }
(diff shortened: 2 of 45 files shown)


DaveDuck321:
Tests fail because I haven't taught the CI how to bootstrap yet. Even after this, some tests are failing because we don't yet have the correct line numbers in the error messages. Everything else should be ready tho...

Time required for a full compilation of the new self-hosted parser:
With lark: 13.38 seconds
With the new parser: 1.73 seconds

Time required to run the full test suite (NOTE: the line number tests fail):
With lark: 16.74 seconds
With the new parser: 3.43 seconds

DaveDuck321:
@antroseco what do you think?
fyi. you can ignore the algorithms in `syntax_tree.c3`: the actual JSON serialization will stay the same but most of the jank there exists purely to emulate the *exact* format that Lark spits out. After this is merged, I'll probably look at completely removing lark to allow the parser to output strongly typed structures, this will hopefully also decrease the complexity of `graphene_parser.py` (which is maybe misleadingly named now that it doesn't do the bulk of the parse).

parser/parser.c3
```diff
@@ -0,0 +1,1381 @@
...
+    // TODO: there must be a nice way to do this
+    let addr : iptr = ref_to_addr(&cursor.program.data[slice.start_offset]);
+    return {&addr_to_heap_array<u8>(addr), slice.end_offset - slice.start_offset};
+}
+
+function slice : (cursor : Cursor&, length : isize) -> ProgramSlice = {
+    return {cursor.position, cursor.position + length};
+}
+
+function slice : (cursor : Cursor&, start : isize, length : isize) -> ProgramSlice = {
```
antroseco, line 26: `cursor` is unused. Is it just for aesthetics?

parser/parser.c3
```diff
@@ -0,0 +1,1381 @@
...
+function get_next_bool_token : (cursor : Cursor&) -> ProgramSlice = {
+    runtime_assert(cursor:is_next("true") | cursor:is_next("false"));
+
+    let start_offset : isize = cursor.position;
+    if cursor:try_skip_next_and_surrounding_white_space("true") {
+        return {start_offset, start_offset + 4};
+    } else {
+        runtime_assert(cursor:try_skip_next_and_surrounding_white_space("false"));
+        return {start_offset, start_offset + 5};
+    }
```
antroseco, line 254: nit: use `"true":length()` instead of 4, same for "false"

antroseco (commented):
Very nice! I'm making my way through `parser.c3` but I'll post the comments for the rest of the code now

parser/parser.c3
```diff
@@ -0,0 +1,1381 @@
...
+
+    // Worst case (not possible without a syntax error)
+    // We've reached the end of the file without seeing a closing triangle bracket
+    brackets_stack:deallocate();
+    return false;
+}
+
+function parse_expression_postfix : (cursor : Cursor&, prefix : Expression) -> Expression = {
+    if cursor:try_skip_next("...") {
+        // TODO: is this right? Do we support expanding expressions?
```
antroseco, line 843: Only in function calls currently; but we could add support for more contexts

parser/parser.c3
```diff
@@ -0,0 +1,1381 @@
...
+        let memory : PackExpansion& = &cursor.allocator:allocate<PackExpansion>();
+        memory = {prefix, meta(prefix.meta.start_offset, cursor:skip_ws_return_offset())};
+        // TODO: Do we support postfixes after expansions?
```
antroseco, line 846: No, and I can't really see how we'd do that

parser/parser.c3
```diff
@@ -0,0 +1,1381 @@
...
+
+        lhs = {8, ref_to_addr(&operator_call), meta};
+        last_operator_start = cursor.position;
+    }
+    return lhs;
+}
+
+function parse_expression : (cursor : Cursor&) -> Expression = {
+    // TODO: this size argument needs to somehow be automatic
+    // TODO: all items on the same line should have the same priority
```
antroseco, line 994: There are algorithms for this https://en.wikipedia.org/wiki/Operator-precedence_parser#Precedence_climbing_method

parser/parser.c3
```diff
@@ -0,0 +1,1381 @@
...
+    };
+    return cursor:parse_expression_with_order_of_operator(order_of_operation:get_iter());
+}
+
+function parse_either_assignment_or_expression : (cursor : Cursor&) -> Line = {
+    let lhs : Expression = cursor:parse_expression();
+    if cursor:try_skip_next_and_surrounding_white_space(";") {
+        // We've reached the end of the line, this must just be an expression
+        let expression_memory : Expression& = &cursor.allocator:allocate<Expression>();
+        expression_memory = lhs;
```
antroseco, line 1015: Perhaps assignments to a & should be more expicit

antroseco (approved):

codegen/interfaces.py
```diff
@@ -275,6 +275,22 @@ def assert_can_write_to(self) -> None:
 GenericMapping = dict[str, SpecializationItem]
 
 
+def do_specializations_match(
```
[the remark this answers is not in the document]
DaveDuck321: The problem is equality with an int triggers the following assertion. I think this is a good sanity check but it breaks when we're mixing int/type specializations.

```python
class Type(ABC):
    def __eq__(self, other: Any) -> bool:
        assert isinstance(other, Type)
		...
```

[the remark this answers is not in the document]
antroseco: Ah sure. I wouldn't mind changing it to `isinstance(other, (Type, int))` but up to you

parser/parser.c3
```diff
@@ -0,0 +1,1381 @@
...
+        runtime_assert(cursor:try_skip_next_and_surrounding_white_space("false"));
+        return {start_offset, start_offset + 5};
+    }
+}
+
+function get_next_decimal_token : (cursor : Cursor&) -> ProgramSlice = {
+    let token : ProgramSlice = cursor:slice(0);
+    if cursor:try_skip_next("-") {
+        token.end_offset += 1;
+    }
```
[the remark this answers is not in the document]
DaveDuck321: In expressions we now parse `-1` as `operator-(1)`.

However, specializations can't have a runtime function call like this, the above code only exists to allow us to parse something like `A<-1>()` (where the `-1` is a signed literal).

This seems a little inconsistent to me so I'd be quite happy to support signed literals again everywhere (alternatively compile time function calls :upside_down_face:).

parser/parser.c3
```diff
@@ -0,0 +1,1381 @@
...
+
+    while !cursor:is_end_of_token() {
+        let next_char : u8 = ascii_to_lower(cursor:get_char());
```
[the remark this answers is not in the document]
DaveDuck321: We don't. Fixed now :partying_face:

parser/parser.c3
```diff
@@ -0,0 +1,1381 @@
...
+}
+
+function parse_named_type : (cursor : Cursor&) -> NamedType = {
+    let name : ProgramSlice = cursor:get_next_ascii_only_identifier();
+    let specialization_list : SpecializationList = cursor:parse_optional_specialization_list();
+    return {name, specialization_list, meta(name.start_offset, specialization_list.meta.end_offset)};
+}
+
+function is_at_start_of_compile_time_constant : (cursor : Cursor&) -> bool = {
+    return cursor:is_next("@") | ascii_is_digit(cursor:get_char()) | cursor:is_next("-");
```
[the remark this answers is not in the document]
DaveDuck321: Yup... Although that probably isn't what we want long-term.

parser/parser.c3
```diff
@@ -0,0 +1,1381 @@
...
+function parse_array_type : (cursor : Cursor&, prefix : Type) -> Type = {
+    runtime_assert(cursor:try_skip_next_and_surrounding_white_space("["));
+
+    let is_heap_array : bool = cursor:try_skip_next_and_surrounding_white_space("&");
+
+    let sizes : Vector<CompileTimeConstant>;
+    // TODO: this would be cleaner with a short-circuit `and`
+    if is_heap_array {
+        // A heap array may/ may not have further dimensions specified
+        // TODO: I'm using `if ... {full} else {allocate empty}` pattern a lot... is there a nicer way
```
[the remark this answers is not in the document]
DaveDuck321: Fair point :grin:

parser/parser.c3
DaveDuck321: While postfixes probably aren't necessary. We might still need prefixes. Let's say you wanted to pass a number of references we might want something like this:
```
function [BigStructs...] fn_name : (others : BigStructs...&) -> void = {
	fn(&others...)
}
```

parser/parser.c3
DaveDuck321: Yeah, possibly... I'm starting to see the value of c pointers.

parser/parser.c3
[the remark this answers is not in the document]
antroseco: I think signed literals are saner overall, although `-(1) != -1` isn't great either

parser/parser.c3
```diff
@@ -0,0 +1,1381 @@
...
+            syntax_error_at_position(&cursor, start_offset, sv("unterminated list, expected '>'"));
+        }
+        if is_missing_comma {
+            syntax_error(&cursor, sv("expected ',' or ']'"));
+        }
+
+        if cursor:is_at_start_of_compile_time_constant() {
+            let constant : CompileTimeConstant& = &cursor.allocator:allocate<CompileTimeConstant>();
+            constant = cursor:parse_compile_time_constant();
+            specializations:push_back<Specialization>({1, ref_to_addr(&constant), meta(start_offset, constant.value.end_offset)});
```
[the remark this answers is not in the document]
antroseco: > you'd need to walk through the entire data structure to deallocate everything

yeah that's was I meant... clearly we need a memory management solution

parser/parser.c3
antroseco: Idk, maybe this would make sense if we try and keep `T&...` consistent with `T&`. `&others...` borrowing everything does make sense though.

Try and guess what happens when you run this though
```lua
@require_once "std/io.c3"

function [T] foo : (a: T) -> void = { puts("T"); }
function [T] foo : (a : T&) -> void = { puts("T&"); }

function main : () -> int = {
    let x : int = 5;
    foo(&x);
    return 0;
}
```

parser/parser.c3
antroseco: lmao

PR #107, opened 2023-08-28
Implement logical operators

antroseco:
This is fairly straightforward, although 708e548a6fed8e8804a16be0302636968a35630d breaks bootstrapping...

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -1,4 +1,19 @@
 #!/usr/bin/env bash
 
-# Atm this is easy... it becomes harder once we remove lark
+set -ex
+
+# Attempt to get the current location. Prefer a branch name if available,
+# otherwise fall back to the current commit's hash if we are in "detached HEAD"
+# state.
+git_location=$(git branch --show)
+if [[ "$git_location" == "" ]]; then
+    git_location=$(git rev-parse HEAD)
+fi
+
+# Stage 1; last commit where the Lark parser can parse the native parser.
+git checkout bootstrap/1
 ./glang ./parser/parser.c3 -o ./parser/parser --bootstrap -O3
+
+# Stage 2; build native parser at current commit
+git checkout "$git_location"
+./glang ./parser/parser.c3 -o ./parser/parser -O3
--- a/parser/parser.c3
+++ b/parser/parser.c3
@@ -978,10 +978,21 @@ function [T] parse_expression_with_order_of_operator : (cursor : Cursor&, op_ite
         let rhs : Expression = cursor:parse_next_operator_subexpression(op_iter);
 
         let meta : Meta = meta(lhs.meta.start_offset, cursor.position);
-        let operator_call : Operator& = &cursor.allocator:allocate<Operator>();
-        operator_call = {op_name, lhs, rhs, meta};
 
-        lhs = {8, ref_to_addr(&operator_call), meta};
+        if is_one_of(cursor:get_sv(op_name), sv("and"), sv("or")) {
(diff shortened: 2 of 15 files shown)


bootstrap.sh
```diff
@@ -1,4 +1,19 @@
...
+    git_location=$(git rev-parse HEAD)
+fi
+
+# Stage 1; last commit where the Lark parser can parse the native parser.
+git checkout bootstrap/1
 ./glang ./parser/parser.c3 -o ./parser/parser --bootstrap -O3
+
+# Stage 2; build native parser at current commit
+git checkout "$git_location"
+./glang ./parser/parser.c3 -o ./parser/parser -O3
```
DaveDuck321, line 19: So after this PR is merged we'll need to iterate through all the commits that break the parser as a number of unique bootstrap stages right?

Will we need to mark each breaking change with a tag? If so can we catch it on CI when we forget to mark a change as breaking

parser/parser.c3
```diff
@@ -24,11 +24,11 @@ function slice : (cursor : Cursor&, length : isize) -> ProgramSlice = {
 }
 
 function ascii_is_text_token_start_char : (char : u8) -> bool = {
-    return ascii_is_alpha(char) | (char == "_"[0]);
+    return ascii_is_alpha(char) or (char == "_"[0]);
 }
 
 function ascii_is_text_token_char : (char : u8) -> bool = {
-    return ascii_is_alphanumeric(char) | (char == "_"[0]);
+    return ascii_is_alphanumeric(char) or (char == "_"[0]);
```
DaveDuck321, line 31: NIT: we can remove these parens now

parser/parser.c3
```diff
@@ -978,10 +978,21 @@ function [T] parse_expression_with_order_of_operator : (cursor : Cursor&, op_ite
...
 
-        lhs = {8, ref_to_addr(&operator_call), meta};
+        if is_one_of(cursor:get_sv(op_name), sv("and"), sv("or")) {
+            // This is currently very similar to the binary operator code.
+            // However, binary operators will be parsed directly as function
+            // calls in the future, so the code will diverge.
+            let logical_operator : LogicalOperator& = &cursor.allocator:allocate<LogicalOperator>();
+            logical_operator = {op_name, lhs, rhs, meta};
+            lhs = {12, ref_to_addr(&logical_operator), meta};
+        }
```
DaveDuck321, line 989: Yeah, fair enough. This all makes sense

DaveDuck321 (approved):
Looks good. Not sure I fully understand the bootstrap strategy but it looks like it'll work for now.

bootstrap.sh
antroseco: Yeah, so I've tagged the commit *before* the breaking change (chained `or` operators in `parser.c3`) as `bootstrap/1`. That's the last commit where the Lark parser can parse the native one. Once we build the native parser, which has proper support for logical operators (you fixed the order-of-operations issue already), we can switch back to a newer version of the native parser, and build that with the stage 1 binary.

To summarise:
- Lark parser used to build old native parser (`bootstrap/1` tag)
- Old native parser (`bootstrap/1` tag) used to build latest parser 

In the future we might also need an extra step at the end:
- Latest parser used to build latest parser

> If so can we catch it on CI when we forget to mark a change as breaking

The CI will always complain if bootstrapping doesn't work, since it bootstraps the compiler every time it runs the test suite.

parser/parser.c3
antroseco: sorry missed these ones

PR #110, opened 2023-08-29
Random clean up and bug fixes

antroseco:


codegen/generatable.py
```diff
@@ -72,8 +72,7 @@ def ir_ref_without_type_annotation(self) -> str:
         assert self.ir_reg is not None
         if isinstance(self.type.definition, CharArrayDefinition):
             return f"@.str.{self.ir_reg}"
-        else:
-            return f"@.{self.ir_reg}"
```
DaveDuck321, line 76: Fine :imp: I'll stop doing this

graphene_parser.py
```diff
@@ -517,6 +521,20 @@ def try_parse_fn_from_specialization(
                 resolved_return = self._program.resolve_type(
                     unresolved_return.produce_specialized_copy(specialization_map)
                 )
+            except SubstitutionFailure:
+                return None  # SFINAE
```
DaveDuck321, line 525: Wait... How did this work before?

std/algorithms.c3
```diff
@@ -36,3 +36,22 @@ function [T] reverse : (buffer : Span<T>&) -> void = {
...
+        if span.data[mid] < elem {
+            lo = mid + 1;
+        }
+        else {
+            hi = mid;
+        }
+    }
+
+    return lo;
+}
```
DaveDuck321, line 57: Amazing!

DaveDuck321 (approved):
Very pretty :heart_eyes:. Our UFCS looks very sexy now

codegen/generatable.py
antroseco: Hahaha, feel free to edit pylintrc

graphene_parser.py
antroseco: We were always catching `SubstitutionFailure`s (now lines 553-554). I had to split the try/catch such that we can attach the location information of the return type specifically.

PR #113, opened 2023-09-04
Add JSON support to the standard library

antroseco:
Adds support for both parsing and dumping JSON.

I've been testing the parser against https://github.com/nst/JSONTestSuite, and it passes most tests except those with string escape sequences and floating point numbers, which I haven't implemented yet.


std/json.c3
```diff
@@ -0,0 +1,428 @@
...
+    return {
+        node_type,
+        make_optional<StringView>(),
+        make_optional<Vector<JSON_Node>>(),
+    };
+}
+
+function char : (literal : u8[1]&) -> u8 = {
+    return literal[0];
+}
```
DaveDuck321, line 36: Ooo, please add this to the standard library. It would also make the parser much cleaner

std/json.c3
```diff
@@ -0,0 +1,428 @@
...
+
+    return token;
+}
+
+function is_json_whitespace : (token : u8) -> bool = {
+    // JSON defines whitespace as one of { space, linefeed, carriage return,
+    // horizontal tab }. If we use string literals instead of code points, then
+    // our code doesn't care about the underlying encoding :).
+    return token == char(" ") or token == char("\n")
+           or token == char("\r") or token == char("\t");
```
DaveDuck321, line 66: `is_one_of(token, ...)`?

std/json.c3
```diff
@@ -0,0 +1,428 @@
...
+}
+
+function eat : (cursor : Cursor&, token : u8) -> bool = {
+    // TODO this should be const (can't use with has_value() right now).
+    let next_char : Optional<u8> = cursor:peak();
+
+    let next_char_data : u8 = next_char:data();
```
DaveDuck321, line 73: Ik its probably fine but we shouldn't really be reading `data` before calling `has_value`. We might want to add a standard library assert in the future.

std/json.c3
```diff
@@ -0,0 +1,428 @@
...
+    for token in string:get_iter() {
+        if !cursor:eat(token) {
+            return false;
+        }
+    }
+
+    return true;
+}
+
+function eat_whitespace : (cursor: Cursor&) -> void = {
```
DaveDuck321, line 92: I kinda like the `eat_*` convention. Might steal that next time I'm working on the parser

std/json.c3
```diff
@@ -0,0 +1,428 @@
...
+        return make_optional<JSON_Node>();
+    }
+
+    let node : JSON_Node = new_json_node(/* object */ 6);
+
+    cursor:eat_whitespace();
+
+    if cursor:eat(char("}")) {
+        // Empty object.
+        return make_optional(node);
```
DaveDuck321, line 118: NIT: imo an length 0 children vector is more ergonomic than a children vector optional since way you can do things like iterate through an object's children without worrying about whether it has children.

Lets also update the vector class to make the empty vector allocation free.

std/json.c3
```diff
@@ -0,0 +1,428 @@
...
+
+    while true {
+        cursor:eat_whitespace();
+
+        let string_node : Optional<JSON_Node> = cursor:parse_string();
+
+        if !string_node:has_value() {
+            node.children:data():deallocate();
+            return make_optional<JSON_Node>();
+        }
```
DaveDuck321, line 132: This is good for now but once we have `map` (or similar) it'd be nice to express that the key can only be a string using the type-system. Then it could just be map<StringView, JSON_Node>

std/json.c3
```diff
@@ -0,0 +1,428 @@
...
+
+        if !cursor:eat(char(":")) {
+            node.children:data():deallocate();
+            return make_optional<JSON_Node>();
+        }
+
+        let value_node : Optional<JSON_Node> = cursor:parse_element();
+
+        if !value_node:has_value() {
+            node.children:data():deallocate();
```
DaveDuck321, line 146: This memory management is a pain. Let's think of something soon

std/json.c3
```diff
@@ -0,0 +1,428 @@
...
+        }
+
+        if !cursor:eat(char(",")) {
+            node.children:data():deallocate();
+            return make_optional<JSON_Node>();
+        }
+    }
+
+    // Unreachable.
+    return make_optional(node);
```
DaveDuck321, line 164: NIT: assert just to be sure :upside_down_face:

std/json.c3
```diff
@@ -0,0 +1,428 @@
...
+
+    cursor:eat_whitespace();
+
+    if cursor:eat(char("]")) {
+        // Empty array.
+        return make_optional(node);
+    }
+
+    // Initialize children vector.
+    node.children:store(cursor.allocator:make_vector<JSON_Node>());
```
DaveDuck321, line 182: Same comment here about the using a length 0 children vector instead of an optional

std/json.c3
```diff
@@ -0,0 +1,428 @@
...
+
+    if next_token == char("[") {
+        return cursor:parse_array();
+    }
+
+    // Numbers don't have a prefix character, so just try and parse them.
+    let node : Optional<JSON_Node> = cursor:parse_number();
+    if node:has_value() {
+        return node;
+    }
```
DaveDuck321, line 232: You can remove this branch by just returning `cursor:parse_number()` (after doing all the other checks ofc)

std/json.c3
```diff
@@ -0,0 +1,428 @@
...
+
+    if !cursor:eat(char("0")) {
+        while is_digit_helper(cursor:peak()) {
+            is_valid = true;
+            cursor:next();
+        }
+    }
+    else {
+        is_valid = true;
+    }
```
DaveDuck321, line 320: Should this be `is_valid=false`? I'm not what you're checking for here

std/json.c3
```diff
@@ -0,0 +1,428 @@
...
+
+    // TODO parse fraction.
+    // TODO parse exponent.
```
DaveDuck321, line 323: I don't think its part of the spec but I've defo also seen `NaN` and `+-Infinity` in JSON files

std/json.c3
```diff
@@ -0,0 +1,428 @@
...
+
+    return make_optional(node);
+}
+
+// TODO this and write below should be the only public functions.
+function json_parse : (source : StringView) -> Optional<JSON_Node> = {
+    // FIXME memory safety, memory leaks.
+    let cursor : Cursor = {
+        source, 0, initialize_allocator()
+    };
```
DaveDuck321, line 342: Lol, leaking the entire allocator is kinda funny tbh. Thats actually the worst possible leak :laughing:

std/json.c3
```diff
@@ -0,0 +1,428 @@
...
+    }
+
+    // false
+    if node.type == 2 {
+        return io:write(sv("false"));
+    }
+
+    // string
+    if node.type == 3 {
+        // Allowed to fail.
```
DaveDuck321, line 379: Why are these allowed to fail?

std/json.c3
```diff
@@ -0,0 +1,428 @@
...
+    if node.type == 6 {
+        // FIXME we should be using a buffer here.
+        io:write(sv("{"));
+        if node.children:has_value() {
+            let iter : VectorIter<JSON_Node> = node.children:data():get_iter();
+            let flag : bool = true;
+
+            // XXX we have to borrow the iterator, otherwise it is not updated
+            // in place by the implicit call to get_next().
+            for element in &iter {
```
DaveDuck321, line 429: Oh damn, smart

DaveDuck321 (approved):
Very nice. I'm amazed you got it done so cleanly without any new standard library.
For the actually exported interface I imagine we'll also need something like:
`function deallocate : (_: JSON_Node&) -> void`

tests/json/LICENSE
```diff
@@ -0,0 +1,21 @@
...
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
```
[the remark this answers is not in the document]
antroseco: Idk, up to you. There's like 320 test files, but we could adapt it to work with our test suite. It'll take a while until we can pass all of them though, so they shouldn't all run on the CI.

std/json.c3
antroseco: Sure, although I still want character literals.

std/json.c3
antroseco: Good shout. It optimized badly unfortunately
```asm
0000000000203ce0 <is_json_whitespace__T_u8>:
  203ce0:	b0 01                	mov    $0x1,%al
  203ce2:	40 80 ff 0a          	cmp    $0xa,%dil
  203ce6:	74 09                	je     203cf1 <is_json_whitespace__T_u8+0x11>
  203ce8:	40 0f b6 cf          	movzbl %dil,%ecx
  203cec:	83 f9 20             	cmp    $0x20,%ecx
  203cef:	75 01                	jne    203cf2 <is_json_whitespace__T_u8+0x12>
  203cf1:	c3                   	ret
  203cf2:	40 80 e7 fb          	and    $0xfb,%dil
  203cf6:	40 80 ff 09          	cmp    $0x9,%dil
  203cfa:	0f 94 c0             	sete   %al
  203cfd:	c3                   	ret
```
vs the uglier code
```asm
0000000000203d50 <is_json_whitespace__T_u8>:
  203d50:	40 80 ff 21          	cmp    $0x21,%dil
  203d54:	0f 92 c1             	setb   %cl
  203d57:	48 b8 00 26 00 00 01 	movabs $0x100002600,%rax
  203d5e:	00 00 00 
  203d61:	48 0f a3 f8          	bt     %rdi,%rax
  203d65:	0f 92 c0             	setb   %al
  203d68:	20 c8                	and    %cl,%al
  203d6a:	c3                   	ret
```

I'll open an issue to follow up on this

std/json.c3
antroseco: Ah I should fix that regardless

> We might want to add a standard library assert in the future.

We could but you can't protect against stuff like this unfortunately

```rust
optional:has_value();
let x : int = optional:data();
```

std/json.c3
antroseco: I stole it from someone else myself :)

std/json.c3
antroseco: I agree, but I didn't want the vectors to allocate every time you made a node.

> Lets also update the vector class to make the empty vector allocation free.

The question then is "what do we return when you call `data()` on an empty vector"? In C++, the standard says you may or may not get a null pointer (that you shouldn't dereference) either way. Are we happy to return an `Optional<Span<T>>`?

std/json.c3
antroseco: That'd be ideal, although I'd rather go with some sort of tree vs a hash map to avoid allocating a bunch of buckets for three elements

std/json.c3
antroseco: Yeah, we really need RAII

std/json.c3
antroseco: Done

std/json.c3
antroseco: Yeah fair enough

std/json.c3
antroseco: The code's correct, sorry it's just a bit confusing. If you check the grammar, https://www.json.org/img/number.png, you'll see that if the first digit is zero, then it can only be followed by the decimal point or the exponent. If you have a non-zero digit, then it can be followed by as many other digits as you want.

The `is_valid` flag is there just to ensure that we've parsed at least one digit. We can't use the cursor cause a single `-` shouldn't count as a valid number.

std/json.c3
antroseco: Yep, they are illegal though :)

std/json.c3
antroseco: Yeah, our memory management is pretty bad :(

antroseco: I really want a global allocator though

std/json.c3
antroseco: Cause we can't return an error! Crashing the whole program sounds pretty excessive tbh

std/json.c3
DaveDuck321: That's a shame... But also wow, that ugly code optimizes amazingly well, llvm is just showing off.

std/json.c3
DaveDuck321: In c++ it's hard because you have pointers. It's easy for us 🙃, you just return a zero length span -- now it's entirely the caller's fault for not checking.

The span's data pointer can just be undefined I don't think it matters, everywhere that takes a span should check there own bounds imo.

std/json.c3
DaveDuck321: I mean:
```
function [T] has_data : (self : Optional<T>&) -> void = {
	if debug {
		runtime_assert(self:has_value());
	}
	...
}
```

std/json.c3
antroseco: Yep, very impressive!

std/json.c3
antroseco: Oh fair. I've added the assert

std/json.c3
antroseco: I guess it's fine as long as you use the Span interface instead of accessing `.data`. Anw, I've changed it

PR #116, opened 2023-09-05
Produce a custom JSON tree rather than the lark format

DaveDuck321:
The parser now produces a typesafe AST rather than an untyped list. This data structure is mirrored in both python and graphene (until we can write the Python binding :-D)

This completely breaks lark, I've removed it (we still require it in the test suites tho).


graphene_parser.py
```diff
@@ -449,185 +279,77 @@ def _parse_generic_function_impl(
...
+        ]
+        unresolved_return = parser.parse_type(node.return_)
+
+        # TODO: this is a temporary hack
+        if len(node.args) > 0 and isinstance(node.args[-1][1], lp.PackType):
+            variadic_args_pack_name, packed_type = node.args.pop()
+
+            # For now a packed type must be a plain type name
+            assert isinstance(packed_type, lp.PackType)
+            assert isinstance(packed_type.type_, lp.NamedType)
```
antroseco, line 309: So a `PackType` is itself a `Type`, but also contains another `Type`, whilst a `Type` is just a `Meta`?

graphene_parser.py
```diff
@@ -740,151 +445,135 @@ def __init__(
...
 
         mapped_value = self._generic_mapping.mapping[argument]
         assert isinstance(mapped_value, int)  # TODO: user facing error
         const_expr = cg.ConstantExpression(cg.IntType(), str(mapped_value))
         return FlattenedExpression([const_expr])
 
-    def SIGNED_INT(self, token: Token) -> FlattenedExpression:
-        const_expr = cg.ConstantExpression(cg.IntType(), token.value)
+    def IntConstant(self, node: lp.IntConstant) -> FlattenedExpression:
+        # TODO: the parser has already decoded this, why are we undoing it?
```
antroseco, line 471: The idea was "why bother if we're just gonna forward it to llvm", but we'll need the actual value soon enough---when I dig up that automatic constant-sizing commit

antroseco (approved):
Nice, super clean!

parser/syntax_tree.c3
```diff
@@ -696,213 +632,272 @@ function write_json_value : (stream : JSON_Stream&, line : Line&) -> void = {
...
+        :end();
 }
 
 function write_json_value : (stream : JSON_Stream&, node : Constant&) -> void = {
-    // TODO: JSON actually supports these types natively
-    //    no need to do this in the compiler (other than lark compat)
+    let object : JSON_Object = stream:new_object();
+    let meta : Meta = get_meta(node.value);
+
+    // JSON supports some types natively, no need to export all of them as strings
```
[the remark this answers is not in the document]
DaveDuck321: Yeah, fair point. It lets us warn on precision loss at the very least.

graphene_parser.py
DaveDuck321: This is the curse and the brilliance of oop :cry: 

- I'm using `Type` as an interface here: anything that can go where you'd normally put a `Type` is a type.
- `Type` like everything the lexer parser produces inherits from `ParsedNode` (which only contains `meta`).
- `PackType` conforms to the `Type` interface because it's used as a type in the function signature. It also contains another `Type` (left of the '...'), atm the only type we support here is a `NamedType` with no specialization hence `NamedType...`. Maybe we want to add references too? Or maybe specializations :upside_down_face: as a way to implement concepts :upside_down_face:?

graphene_parser.py
DaveDuck321: Ooo, yes I was missing this, that would be cool.

I gave up on the Unicode library partially because it was too painful to work without automatic const-sizing + char literals (also because I released the parser didn't actually need it :smile:).

graphene_parser.py
antroseco: Ok that makes sense I guess

graphene_parser.py
antroseco: Fair enough. We technically do need a unicode library to properly do json though :/

bootstrap.sh
```diff
@@ -11,8 +11,8 @@ function checkout() {
 stage_1_dir=./bootstrap-1
 stage_2_dir=./bootstrap-2
 
-checkout bootstrap/2 $stage_2_dir
 checkout bootstrap/1 $stage_1_dir
+checkout bootstrap/2 $stage_2_dir
```
antroseco, line 15: that did bother me ngl.. but I didn't want to be tooo annoying :)

DaveDuck321: :stuck_out_tongue_closed_eyes:

PR #117, opened 2023-09-05
Package project with hatch

antroseco:
This is not ready yet, but let me know what you think.
Hatch seems to be the modern way to manage python projects, so I thought we'd try that.

The workflow is quite different though. You use `hatch env` and `hatch shell` instead of sourcing scripts from `venv`. I configured three different environments instead of one: default is just the runtime dependencies, test has the test dependenceis as well, and style is only black and isort. Each environment has scripts configured, so you can test with `hatch run test:run` and check formatting with `hatch run style:check`.

The main benefit of restructuring the project like this (and having the `src` directory), is that the project is *installed* in the environments, so if stuff like imports work on our machines then they should work everywhere.


pyproject.toml
```diff
@@ -0,0 +1,53 @@
...
+
+[project.urls]
+"Homepage" = "https://github.com/DaveDuck321/GrapheneLang/"
+"Bug Tracker" = "https://github.com/DaveDuck321/GrapheneLang/issues"
+
+[project.scripts]
+glang = "glang.driver:main"
+
+[tool.hatch.envs.test]
+dependencies = ["interegular==0.3.2", "lark==1.1.7"]
```
antroseco, line 32: This inherits from default

pyproject.toml
```diff
@@ -0,0 +1,53 @@
...
+
+[tool.hatch.envs.style.scripts]
+check = [
+  "black --check --diff .",
+  "isort --check-only --diff --profile black .",
+]
+fmt = ["isort --profile black .", "black ."]
+
+[tool.hatch.build.targets.sdist]
+exclude = [".github", "docs", "tests"]
```
antroseco, line 50: Excludes some stuff from the source distribution

pyproject.toml
```diff
@@ -0,0 +1,53 @@
...
+
+[tool.hatch.build.targets.wheel]
+packages = ["src/glang"]
```
antroseco, line 53: What pip actually installs. I still need to include the parser binary and the `lib` directory

antroseco:
You'll need to install hatch's arch package if you want to play around with this https://hatch.pypa.io/latest/install/

antroseco:
Was a pain to set up, but I think everything works now

glang.spec
```diff
@@ -0,0 +1,61 @@
+Name:           glang
+Version:        0.1.3
+Release:        %autorelease
+Summary:        A Graphene front-end for LLVM
+
+License:        AGPL-3.0-only
+URL:            https://github.com/DaveDuck321/GrapheneLang
+Source0:        %{url}/archive/refs/heads/dev_hatch.tar.gz
```
DaveDuck321, line 8: How are we going to keep this updated? Do we just tag releases and update the URL in this file?

Also is this python ecosystem or Fedora-specific?

pyproject.toml
```diff
@@ -0,0 +1,82 @@
...
+
+[project]
+name = "glang"
+version = "0.1.3"
+authors = [
+  { name = "Antros Economides", email = "antroseconomides@hotmail.co.uk" },
+  { name = "Tom Grant", email = "thomas.grant.mail@gmail.com" },
+]
+description = "A Graphene front-end for LLVM"
+# TODO readme = "README.md"
```
DaveDuck321, line 14: lol, as if

pyproject.toml
```diff
@@ -0,0 +1,82 @@
...
+license = "AGPL-3.0-only"
+requires-python = ">=3.10"
+classifiers = [
+  "Development Status :: 3 - Alpha",
+  "License :: OSI Approved :: GNU Affero General Public License v3",
+  "Operating System :: POSIX :: Linux",
+  "Programming Language :: Python :: 3",
+  "Topic :: Software Development :: Compilers",
+]
+dependencies = ["PyYAML~=6.0.0", "typed-argument-parser~=1.8.1"]
```
DaveDuck321, line 24: Is there also a `provides = "glang"` field?

pyproject.toml
```diff
@@ -0,0 +1,82 @@
...
+
+[tool.hatch.envs.test.scripts]
+run = "python tests/run_tests.py"
+
+
+# Style environment. Does not depend on the project.
+[tool.hatch.envs.style]
+detached = true
+skip-install = true
+dependencies = ["black", "isort"]
```
DaveDuck321, line 55: `pylint` also?

tests/run_tests.py
```diff
@@ -220,18 +221,20 @@ def run_tests(tests: list[Path], workers: int) -> int:
...
-    if (
-        not RUNTIME_OBJ_PATH.is_file()
-        or runtime_src_path.stat().st_mtime > RUNTIME_OBJ_PATH.stat().st_mtime
-    ):
-        # Need to use `cc` because `as` doesn't run the C preprocessor.
-        subprocess.run(
-            ["cc", "-c", runtime_src_path, "-o", RUNTIME_OBJ_PATH], check=True
-        )
+    with resources.as_file(
+        resources.files("glang.lib.std") / HOST_TARGET / "runtime.S"
```
DaveDuck321, line 225: Huh, cool. I've never seen this... This looks like something from native app development not python.

glang.spec
```diff
@@ -0,0 +1,61 @@
+Name:           glang
+Version:        0.1.3
```
DaveDuck321, line 2: Where does this number come from :upside_down_face:?

DaveDuck321 (commented):
Ok, this is pretty cool. I tried it out locally and everything seemed to work well. It makes the AUR package easier too :smile:.

~~I see you've disabled the vscode venv terminal integration., I'm assuming this is so that hatch manages its own state. The LSP doesn't seem to recognize hatch's enviroment tho, should I just point vscode to the venv that hatch creates in the local usr folder (while the terminal is still disconnected)?~~

hatch_build.py
```diff
@@ -0,0 +1,31 @@
...
+        Platform().check_command(["./bootstrap.sh"], shell=False)
+
+        self.app.display_success("Done!")
+
+        build_data["force_include"]["dist/parser"] = "glang/bin/parser"
+
+        build_data["pure_python"] = False
+        build_data["infer_tag"] = True
+
+        return super().initialize(version, build_data)
```
DaveDuck321, line 31: Is this only run once when I create the environment? Let's say I wanted to redo the bootstrap... would I do it manually, rebuild hatch, or something else entirely?

pyproject.toml
```diff
@@ -0,0 +1,82 @@
...
+"Bug Tracker" = "https://github.com/DaveDuck321/GrapheneLang/issues"
+
+[project.scripts]
+glang = "glang.driver:main"
+
+
+# Development environment to be used with your IDE. Just set the venv path to
+# the output of `hatch env find dev`.
+[tool.hatch.envs.dev]
+dependencies = ["hatch~=1.7.0", "lark~=1.1.7"]
```
DaveDuck321, line 37: Oh I see. That answers one of my questions then.

glang.spec
antroseco: > How are we going to keep this updated? Do we just tag releases and update the URL in this file?

Yeah we can tag releases and have the URL update automatically with the `%{version}` macro. Eventually I'd like to move the version to a single place and have the packaging files just read it from there.

We can also build on every commit if we want, instead of releases.

Another option was to have the spec file in its own repo (which is what distro package maintainers normally do), but then you have to manually update it on every release. 

> Also is this python ecosystem or Fedora-specific?

The spec file is just for RPMs. So Fedora and RHEL (+clones), as well as SUSE.

pyproject.toml
antroseco: one day :)

pyproject.toml
antroseco: I don't think so. If you require `glang` from another package it should just work

glang.spec
antroseco: I started at `0.1.0` and incremented the patch every time I wanted to push to `pypi` :)

pyproject.toml
antroseco: Yeah, the `dev` environment exists just for vscode

hatch_build.py
antroseco: Yes. It's only run when you build the package (which you implicitly do the first time you create an environment). You can also run `hatch build` to manually rebuild, or just `hatch run dev:bootstrap` to only run `bootstrap.sh` in the `dev` environment (this is defined in `pyproject.toml`.

antroseco:
We have RPMs at https://copr.fedorainfracloud.org/coprs/antros/graphene/packages/ and wheels on PyPI at https://pypi.org/project/glang/ now

DaveDuck321 (approved):

pyproject.toml
antroseco: Right, good catch. I ended up adding it to the `dev` environment since it checks whether all module imports are valid.

tests/run_tests.py
antroseco: Yeah... I really wanted this to work so that we can split our source tree, but it doesn't really do what I wanted... It's more effort than it's worth

Issue #25, opened 2022-12-31
Throw specialization errors if generic is not matched 


DaveDuck321:
```c++
typedef[T] s: i64;
typedef s<>: i32;
typedef s<int, int>: i32;
```
This is currently valid. We should ensure specializations take the same number of types as the definition.

DaveDuck321:
As of #111 we explicitly allow this in order to support parameter packs

DaveDuck321:
> As of https://github.com/DaveDuck321/GrapheneLang/pull/111 we explicitly allow this in order to support parameter packs

I no longer agree with my opinion here due to the following code:
```
typedef [T] Hello : {}
typedef [T] Hello<int> : {}

Hello<int> // Instantiates the first typedef
```

Did you spot the typo? The second `Hello` is actually of type `Hello<int, T>` due to its generic argument. This error is silent and difficult to detect.

We'll need to work out something for packed types tho.... Maybe we just don't allow packed types and instead, we use functions to instantiate them?

antroseco:
> > As of #111 we explicitly allow this in order to support parameter packs
> 
> I no longer agree with my opinion here due to the following code:
> 
> ```
> typedef [T] Hello : {}
> typedef [T] Hello<int> : {}
> 
> Hello<int> // Instantiates the first typedef
> ```
> 
> Did you spot the typo? The second `Hello` is actually of type `Hello<int, T>` due to its generic argument. This error is silent and difficult to detect.

Ngl that's what I thought would happen in this case, but I can see why this is confusing.

> We'll need to work out something for packed types tho.... Maybe we just don't allow packed types and instead use functions to instantiate them?

```
typedef [Ts...] Hello : {}
typedef Hello<int> : {}
typedef Hello<int, int> : {}
```

Maybe this should be allowed. It looks fine to me as long as you disallow any other definitions with generics like:

```
typedef [T] Hello<int> : {}
```

Actually, what if we enforce this syntax?

```
typedef [T] Hello<int, T> : {}
```

DaveDuck321:
> Actually, what if we enforce this syntax?

Ooo, yeah, this seems reasonable.

Let's say we ONLY automatically generate the default specialization if no specialization has been provided. Otherwise, we error if not all generic definitions are present in the specialization.

PR #118, opened 2023-09-08
Allow specializations with generic arguments

DaveDuck321:
This turns on partial-type specializations and adds some error messages. I give an example showing how the `Optional` type can be specialized to optimize optional references.

Also as suggested in #25 we now error when the specialization count is mismatched (see the issue for the reasoning behind my flip-flop).


codegen/type_resolution.py
```diff
@@ -1091,6 +1115,19 @@ def lookup_named_type(
                 f"typedef {prefix}{format_specialization(specialization)} : ...",
             )
 
+        example_candidate = self._unresolved_typedefs[prefix][0]
```
antroseco, line 1158: Is this guaranteed to exist?

tests/regressions/error_in_return_type.c3
```diff
@@ -1,8 +1,6 @@
 typedef [T] Span : {}
 
-// Hmm, the error is a bit misleading...
-// The overload resolution fails due to SFINAE not parameter count.. but this is
-// also kinda the point of SFINAE, if we fixed this `EnableIf` would break.
+// TODO: the line number still isn't right here
 function [T] foo : ()
     -> Span = {}
```
antroseco, line 5: I swear I fixed that in a previous commit, before your big refactor. I just added a separate try/catch block around the code that parses the return type, so that it would wrap the error with the return type's location info.

tests/generic_types/duplicate_type_specialization.c3
```diff
@@ -7,4 +7,4 @@ typedef MyType<u8> : int
 /// File *.c3', line 3, in 'typedef'
 /// Error: multiple definitions of type 'MyType<u8>':
 /// - typedef MyType<u8> : bool      (File '*.c3', line 3)
-/// - typedef MyType<u8> : bool      (File '*.c3', line 4)
+/// - typedef MyType<u8> : int       (File '*.c3', line 4)
```
antroseco, line 10: Oh damn... Can't believe I missed that

std/type_traits.c3
```diff
@@ -15,6 +15,9 @@ typedef Either<FalseType, FalseType> : FalseType
 typedef Not<TrueType> : FalseType
 typedef Not<FalseType> : TrueType
 
+typedef [U, V] AreEquivalent : FalseType
+typedef [V] AreEquivalent<V, V> : TrueType
```
antroseco, line 21: Finally!!! I've wanted this for ages

std/arithmetic.c3
```diff
@@ -15,12 +15,7 @@ typedef CommonArithmeticType<i32, i128> : i128
 typedef CommonArithmeticType<i64, i128> : i128
 
 typedef [U, V] ArithmeticCompareResult : TypeIf<bool, Both<IsIntegral<U>, IsIntegral<V>>>
-
-typedef HasTrivialAssignments<i8> : TrueType
-typedef HasTrivialAssignments<i16> : TrueType
-typedef HasTrivialAssignments<i32> : TrueType
-typedef HasTrivialAssignments<i64> : TrueType
-typedef HasTrivialAssignments<i128> : TrueType
+typedef [T] RegisterHasTrivialAssignments<Overwrite<T>> : TypeIf<TrueType, IsIntegral<T>>
```
antroseco, line 18: What's the purpose of `Overwrite`?

codegen/type_resolution.py
```diff
@@ -1038,12 +1040,27 @@ def select_function_with_least_implicit_conversion_cost(
...
+        for other in self._already_codegened[declaration.uuid]:
+            if (
+                do_specializations_match(other.specialization, signature.specialization)
+                and other.arguments == signature.arguments
+                and other.return_type == signature.return_type
+            ):
+                return  # Already codegened
+
+        self._already_codegened[declaration.uuid].append(signature)
+        self._remaining_to_codegen.append((declaration, signature))
```
antroseco, line 1055: Perhaps `_already_codegened` should be a set then

codegen/type_resolution.py
```diff
@@ -1038,12 +1040,27 @@ def select_function_with_least_implicit_conversion_cost(
...
+
     def lookup_function(
         self,
         name: str,
         given_specialization: list[SpecializationItem],
         args: list[TypedExpression] | list[Type],
-    ) -> FunctionSignature:
+        evaluated_context: bool,
```
antroseco, line 1062: Is this always `true`?

antroseco (approved):
Very cool!

codegen/type_resolution.py
DaveDuck321: Yup. The previous lines confirm that ```prefix in self._unresolved_typedefs```. Ofc the python type system implies `self._unresolved_typedefs` can be an empty list, but this never happens since we build `self._unresolved_typedefs` by appending typedefs.

tests/generic_types/duplicate_type_specialization.c3
DaveDuck321: :cry: me too. Noticing this would have saved a lot of debugging pain

codegen/type_resolution.py
DaveDuck321: Yeah, this would be a valid optimization. Atm it would be kinda annoying to freeze the signature and add the hashing/ equality comparisons so its linear complexity for now :upside_down_face:.

codegen/type_resolution.py
DaveDuck321: At the moment yes. It will only be false when we add something like `decltype`.

codegen/type_resolution.py
antroseco: Let's do it at some point though. Just changing `do_specializations_match()` to list equality is a 5.5% performance boost, if we do even fewer comparisons by caching hashes etc then it'll add up very quickly

codegen/type_resolution.py
antroseco: Oh I see

std/arithmetic.c3
DaveDuck321: Let's say we want to do something like this:
```c++
typedef [T] MyTypeTrait : FalseType // The default case
typedef [T] MyTypeTrait : TypeIf<TrueType, IsIntegral<T>>  // Integral types should all have `MyTypeTrait`
```
`MyTypeTrait<int>` fails with the error `'MyTypeTrait<int>' is ambiguous` since both typedefs pass SFINAE and have the same pattern matching cost.

`Override` (which I now think is a better name than `Overwrite`) gets around this issue by giving the default case a higher pattern maching cost than the specialized versions:
```c++
typedef [T] RegisterMyTypeTrait<Override<T>> : ... // Pattern match `T` in `Override<T>`
typedef [T] RegisterMyTypeTrait<T> : ...  // Pattern match `T` by itself
```

antroseco: Ahhh, nice. That makes sense

tests/regressions/error_in_return_type.c3
DaveDuck321: Yeah, this defo worked perfectly before the refactor.

I introduced this problem by allowing functions to sit around in an unresolved state. This allows us to resolve everything simultaneously but means that we only know about these errors after the program tries to resolve an exact function signature (by this point in time we've lost the exact line numbers).

I've made this much better (but not perfect) in e49243b608b8044df3c5ef6f5b52022b46815f21 by wrapping any errors with the line number of the function signature (note: in this example that means the line number is off-by-one).

antroseco: Right. We don't resolve inside `graphene_parser.py` anymore. I guess we'll have to start passing the tokens too.

PR #120, opened 2023-09-13
Add red-black trees to the standard library

antroseco:
Required for an efficient `map` container.

Insertion and search work, but we are missing ~~deletion and~~ the traversals.

### DIFF
--- a/src/glang/lib/std/rb_tree.c3
+++ b/src/glang/lib/std/rb_tree.c3
@@ -424,7 +424,7 @@ function [K] delete_impl : (self : RB_Tree<K>&, u : RB_Node<K>&) -> void = {
     if u.left:has_value() and u.right:has_value() {
         // u has two children.
         let v : RB_Node<K>& = &u.right:data():minmax(LEFT());
-        must_fix = v.colour == BLACK();
+        must_fix = (v.colour == BLACK());
         x = v.right;
 
         if &v != &u.right:data() {
--- a/src/glang/lib/std/rb_tree.c3
+++ b/src/glang/lib/std/rb_tree.c3
@@ -431,7 +431,7 @@ function [K] delete_impl : (self : RB_Tree<K>&, u : RB_Node<K>&) -> void = {
             // v doesn't have a left child, otherwise that child would be the
             // minimum. It might have a right child though, so replace v with
             // its right child before we move it around the tree. Set x_parent
-            // before the shift, as v might be an empty optional.
+            // before the shift, as the right child might be an empty optional.
             x_parent = v.parent;
             self:replace_node_with(&v, v.right);
 
--- a/src/glang/lib/std/rb_tree.c3
+++ b/src/glang/lib/std/rb_tree.c3
@@ -128,9 +128,9 @@ function [K, V] rotate : (
     let v : RB_Node<K, V>& = &get_child(&u, !side):data();
 
     // Move child from v to u, and update the child's parent pointer.
-    get_child(&u, !side) = get_child(&v, side);
-    if get_child(&u, !side):has_value() {
-        get_child(&u, !side):data().parent:store(&u);
+    u:get_child(!side) = v:get_child(side);
+    if u:get_child(!side):has_value() {
+        u:get_child(!side):data().parent:store(&u);
     }
 
     // Replace u with v.
--- a/src/glang/lib/std/rb_tree.c3
+++ b/src/glang/lib/std/rb_tree.c3
@@ -422,7 +422,7 @@ function [K] delete_impl : (self : RB_Tree<K>&, u : RB_Node<K>&) -> void = {
     // which is represented as an empty optional.
     let x : Optional<RB_Node<K>&> = make_optional<RB_Node<K>&>();
     let x_parent : Optional<RB_Node<K>&>;
-    let must_fix : bool = u.colour == BLACK();
+    let must_fix : bool = (u.colour == BLACK());
 
     if u.left:has_value() and u.right:has_value() {
         // u has two children.
--- a/src/glang/lib/std/rb_tree.c3
+++ b/src/glang/lib/std/rb_tree.c3
@@ -17,6 +17,14 @@ function RIGHT : () -> bool = {
     return true;
 }
 
+function left_if : (condition : bool) -> bool = {
+    if condition {
+        return LEFT();
+    }
+
+    return RIGHT();
+}
+
 typedef [K, V] RB_Node : {
     key    : K,
     value  : V,
@@ -231,28 +239,18 @@ function [K, V] insert : (
             u.value = value;
             return;
         }
-        else {
-            // Keep track of the parent node.
-            p_opt:store(&u_ptr:data():data());
 
-            if key < u.key {
-                u_ptr = {&u.left};
-            }
-            else {
-                u_ptr = {&u.right};
-            }
-        }
+        // Keep track of the parent node.
+        p_opt:store(&u_ptr:data():data());
+
+        u_ptr = {&u:get_child(left_if(key < u.key))};
     }
 
     u_ptr:data():store(&self:new<RB_Node<K, V>>(key, value, p_opt));
 
     if p_opt:has_value() {
-        if key < p_opt:data().key {
-            p_opt:data().left = u_ptr:data();
-        }
-        else {
-            p_opt:data().right = u_ptr:data();
-        }
+        let p : RB_Node<K, V>& = &p_opt:data();
+        p:get_child(left_if(key < p.key)) = u_ptr:data();
     }
 
     self:fix_after_insert(&u_ptr:data():data());
@@ -268,12 +266,7 @@ function [K, V] search : (self : RB_Tree<K, V>&, key : K) -> Optional<V> = {
             return make_optional(u:data().value);
         }
 
-        if key < u_key {
-            u = u:data().left;
-        }
-        else {
-            u = u:data().right;
-        }
+        u = u:data():get_child(left_if(key < u_key));
     }
 
     return make_optional<V>();
--- a/src/glang/lib/std/rb_tree.c3
+++ b/src/glang/lib/std/rb_tree.c3
@@ -42,22 +42,24 @@ typedef [K] RB_Tree : {
     return ref_to_addr(&left) == ref_to_addr(&right);
 }
 
-function [IO, K] write : (
-    io : IO&, node : RB_Node<K>
-) -> TypeIf<StringView, IsStreamableTextIO<IO>> = {
-    // NOTE for debugging only.
-    io:write(node.key);
-    if node.colour == RED() {
-        io:write(sv(" (red)"));
+@operator [K] != : (left : RB_Node<K>&, right : RB_Node<K>&) -> bool = {
+    return !(&left == &right);
+}
+
+@operator [K] == : (
+    left : Optional<RB_Node<K>&>, right : Optional<RB_Node<K>&>
+) -> bool = {
+    if left:has_value() and right:has_value() {
+        return &left:data() == &right:data();
     }
-    else {
-        io:write(sv(" (black)"));
-    }
-    io:write(sv("\nleft: {\n"));
-    io:write(node.left);
-    io:write(sv("\n}\nright: {\n"));
-    io:write(node.right);
-    return io:write(sv("\n}\n"));
+
+    return left:has_value() == right:has_value();
+}
+
+@operator [K] != : (
+    left : Optional<RB_Node<K>&>, right : Optional<RB_Node<K>&>
+) -> bool = {
+    return !(left == right);
 }
 
 function [K] new<RB_Node<K>> : (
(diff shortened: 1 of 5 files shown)


src/glang/lib/std/rb_tree.c3
```diff
@@ -0,0 +1,273 @@
...
+    if condition {
+        return LEFT();
+    }
+
+    return RIGHT();
+}
+
+typedef [K, V] RB_Node : {
+    key    : K,
+    value  : V,
```
DaveDuck321, line 30: Interestingly llvm/ gcc's implementations don't have a key-value pair in each node, instead, they only store a single piece of data. So for example to implement a map you would specialize it as `RB_Node<MapData<K, V>>` (where `MapData = {key : K, value : V}`). This would also make ordered sets trivial `RB_Node<K>`.

For this to work, you'd need to adjust the interface slightly since we can't use the less-than comparison. Instead of `insert(tree, key, value)`, you could have `insert_in_position(tree, predecessor, data)` called by wrappers in `map` and `set` respectively. This has the interesting consequence that the red-black tree doesn't actually perform any comparisons of its nodes, it just preserves the ordering you give it... so maybe we could store something that can't be ordered with comparisons :scream:, not sure what tho :upside_down_face:.

src/glang/lib/std/rb_tree.c3
```diff
@@ -0,0 +1,273 @@
...
+    left   : Optional<RB_Node<K, V>&>,
+    right  : Optional<RB_Node<K, V>&>,
+    parent : Optional<RB_Node<K, V>&>,
+    colour : bool
```
DaveDuck321, line 34: :gb:

src/glang/lib/std/rb_tree.c3
```diff
@@ -0,0 +1,273 @@
...
+
+typedef [K, V] RB_Tree : {
+    root : Optional<RB_Node<K, V>&>,
+    allocator : Allocator,  // FIXME use global allocator.
+}
+
+@operator [K, V] == : (left : RB_Node<K, V>&, right : RB_Node<K, V>&) -> bool = {
+    // Assume both nodes belong to the same tree, to keep things simple.
+    return left.key == right.key;
+}
```
DaveDuck321, line 45: Let's just compare the addresses here?

src/glang/lib/std/rb_tree.c3
```diff
@@ -0,0 +1,273 @@
...
+
+        // Case 2: u has a red parent and a red uncle.
+        if w.left:is_red() and w.right:is_red() {
+            w.left:colour_black();
+            w.right:colour_black();
+
+            // Push the issue up the tree.
+            w.colour = RED();
+            u_ptr = {&w};
+            // TODO continue;
```
DaveDuck321, line 197: Basic fundamental language feature :upside_down_face:

src/glang/lib/std/rb_tree.c3
```diff
@@ -0,0 +1,273 @@
...
+
+    u_ptr:data():store(&self:new<RB_Node<K, V>>(key, value, p_opt));
+
+    if p_opt:has_value() {
+        let p : RB_Node<K, V>& = &p_opt:data();
+        p:get_child(left_if(key < p.key)) = u_ptr:data();
+    }
+
+    self:fix_after_insert(&u_ptr:data():data());
+}
```
DaveDuck321, line 257: I think this should be part of the `map`s interface not necessarily a general purpose red-black tree

DaveDuck321 (commented):
This is very cool, I like what you've done so far.

I think it'd be worth factoring `map` functionality out so we have a pure `rb_tree` but your actual implementation is very clean regardless.

src/glang/lib/std/rb_tree.c3
```diff
@@ -0,0 +1,273 @@
...
+    return true;
+}
+
+function LEFT : () -> bool = {
+    return false;
+}
+
+function RIGHT : () -> bool = {
+    return true;
+}
```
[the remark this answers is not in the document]
antroseco: Yep

src/glang/lib/std/rb_tree.c3
antroseco: lmao

src/glang/lib/std/rb_tree.c3
antroseco: Done

src/glang/lib/std/rb_tree.c3
```diff
@@ -0,0 +1,273 @@
...
+    u:get_child(!side) = v:get_child(side);
+    if u:get_child(!side):has_value() {
+        u:get_child(!side):data().parent:store(&u);
+    }
+
+    // Replace u with v.
+    v.parent = u.parent;
+    if v.parent:has_value() {
+        const u_side : bool = which_child_am_i(&u);
+        u.parent:data():get_child(u_side):store(&v);
```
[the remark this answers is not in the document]
antroseco: Perhaps. It looks kidna weird followed by the . or :, but I'm not sure if there's much else we can do. Fyi, Cpp2 uses `x*`, so `x*.y`.

antroseco:
We have a working `delete` implementation now. I know it's hard, but it'd be great if you could review it super thoroughly---I had to deviate from the notes and re-derive some parts of the algorithm.

src/glang/lib/std/rb_tree.c3
```diff
@@ -0,0 +1,481 @@
...
+}
+
+function [K] fix_after_delete_impl : (
+    self : RB_Tree<K>&, x : Optional<RB_Node<K>&>, y : RB_Node<K>&
+) -> Optional<RB_Node<K>&> = {
+    // x is guaranteed not to be the root. y is guaranteed to be x's parent.
+    let x_side : bool = which_child_am_i(x, &y);
+    let w_side : bool = !x_side;
+
+    // x's sibling, w, cannot be a leaf node (nil node). If x is double black,
```
DaveDuck321, line 331: ~~Why can't both `x` and `w` be nil leaf nodes if `x` is "black-and-red"?~~
Ohh, leaf nodes can only ever be double black right?

src/glang/lib/std/rb_tree.c3
```diff
@@ -0,0 +1,481 @@
...
+
+    // We have removed one black node from w's side of the tree (w itself), so
+    // we need to make its red child black to compensate (as it is now the root
+    // of that subtree). On the other side, the new root should also be black,
+    // as it absorbs the "double black" from x.
+    w.left:colour_black();
+    w.right:colour_black();
+
+    // Remove the double black; we are done.
+    return self.root;
```
DaveDuck321, line 388: Wow, the rest of it was quite easy but this took some thinking. Can confirm, this works in my head (and on my tablet).

DaveDuck321 (approved):
Very cool.

I've gone through the algorithm and it looks good. I didn't verify the lecture note were actually correct but I can confirm you've matches all the cases in the same way Sam did. I did confirm the algorithm were correct when you've deviated tho.

This is defo the most sophisticated algorithm that been written in graphene (can't wait for the JSON library support :upside_down_face: (whoever ends up adding it))

src/glang/lib/std/rb_tree.c3
antroseco: Yeah, the implicit assumption in all of this is that you start with a valid red-black tree, so leaf nodes are always (originally) black :)

DaveDuck321:
Damn, this would be kinda useful rn. I'll just merge it next time I need it... Is anything obviously broken?

src/glang/lib/std/rb_tree.c3
antroseco: I think this is sensible, but we have an opportunity here to avoid a mistake (IMHO) in the C++ standard library and enable heterogeneous lookups by default. This lets us keep the `insert` function, but rely on an operator < template that we'll define, which will call the user-defined operator < for the key.

Not sure if there are any obvious downsides to this approach though

src/glang/lib/std/rb_tree.c3
```diff
@@ -0,0 +1,481 @@
...
+    x:colour_black();
+}
+
+function [K] delete_impl : (self : RB_Tree<K>&, u : RB_Node<K>&) -> void = {
+    // x is the "double black" or "red-and-black" node which we use to fix any
+    // red-black tree violations that arise from the deletion. y is its parent,
+    // as x might be a leaf node (nil node), which is represented as an empty
+    // optional.
+    let x : Optional<RB_Node<K>&> = make_optional<RB_Node<K>&>();
+    let y : Optional<RB_Node<K>&>;
```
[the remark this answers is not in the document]
antroseco: Done

src/glang/lib/std/rb_tree.c3
```diff
@@ -0,0 +1,481 @@
...
+    let must_fix : bool = u.colour == BLACK();
```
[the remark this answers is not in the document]
antroseco: Sure, done

tests/stdlib/rb_tree/extended.c3
```diff
@@ -0,0 +1,128 @@
...
+    if !child_opt:has_value() {
+        return make_optional(1);
+    }
+
+    let child : RB_Node<K>& = &child_opt:data();
+
+    if side == LEFT() {
+        if child.key >= node.key {
+            return make_optional<int>();
+        }
```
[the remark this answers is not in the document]
antroseco: Hmm, you're right, but I'd rather make the test as strict as possible

src/glang/lib/std/rb_tree.c3
```diff
@@ -0,0 +1,481 @@
...
+    let u : Optional<RB_Node<K>&> = self.root;
+
+    while u:has_value() and key != u:data().key {
+        u = u:data():get_child(left_if(key < u:data().key));
+    }
+
+    return u;
+}
+
+function [K] search : (self : RB_Tree<K>&, key : K) -> Optional<K> = {
```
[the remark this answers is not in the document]
antroseco: Yeah, we should definitely allow the `key` here to be of a different type. The problem though is that we can't (currently) bind a temporary to a reference, so `rb_tree:search(5)` no longer works. I'll open an issue for this.

src/glang/lib/std/rb_tree.c3
```diff
@@ -0,0 +1,481 @@
...
+        let side : bool = left_if(u.left:has_value());
+        x = u:get_child(side);
+        y = u.parent;
+
+        // And replace u with it. It's fine if x is empty; the code is just
+        // simpler this way.
+        self:shift_nodes(&u, x);
+    }
+
+    self.allocator:deallocate(&u);
```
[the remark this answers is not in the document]
antroseco: Ikr...

src/glang/lib/std/rb_tree.c3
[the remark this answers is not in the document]
antroseco: I've made it an empty optional, although that doesn't really do much for us, because what we really want is to check that it has been assigned to. Did you have anything else in mind?

src/glang/lib/std/rb_tree.c3
```diff
@@ -0,0 +1,481 @@
...
+        // u has two children.
+        let v : RB_Node<K>& = &u.right:data():minmax(LEFT());
+        must_fix = v.colour == BLACK();
+        x = v.right;
+
+        if &v != &u.right:data() {
+            // v doesn't have a left child, otherwise that child would be the
+            // minimum. It might have a right child though, so replace v with
+            // its right child before we move it around the tree. Set y before
+            // the shift, as v might be an empty optional.
```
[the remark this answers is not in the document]
antroseco: Nice spot. Pretty sure I meant "v's right child" instead (which can be an empty optional)

antroseco:
That was... actually painful. Depends on #135 but otherwise ready to merge.

DaveDuck321:
Looks like this is pulling some commits from #135. I'll take another look when that's merged.

antroseco:
> Looks like this is pulling some commits from #135. I'll take another look when that's merged.

The diff looked wrong even after merging #135 into main. I ended up rebasing and now it's clean.

src/glang/lib/std/wrappers.c3
```diff
@@ -55,9 +65,20 @@ function [T] has_value : (self : Optional<T>&) -> bool = {
...
+        return io:write(optional:data());
+    }
+
+    return io:write("<empty>");
+}
+
+
 // Optional<T&>
-function [T] make<T&> : (value : T&) -> Optional<T&> = {
-    return { ref_to_addr(&value) };
```
DaveDuck321, line 60: Oh wow, this was a bug. Good catch

src/glang/lib/std/wrappers.c3
```diff
@@ -92,11 +114,11 @@ function [T] has_value : (self : Optional<T&> mut&) -> bool = {
...
-function [T] make<T&> : (value : T mut&) -> Optional<T mut&> = {
-    return { ref_to_addr(&value) };
+function [T] make<Optional<T mut&>> : () -> Optional<T mut&> = {
+    return { 0 };
 }
 
-function [T] make<T&> : (value : T mut&) -> Optional<T mut&> = {
+function [T] make<T mut&> : (value : T mut&) -> Optional<T mut&> = {
     return { ref_to_addr(&value) };
 }
```
DaveDuck321, line 123: This has the same bug you fixed in the immutable case eg. it should be `make<Optional<T mut&>>`

DaveDuck321 (approved):
Beautiful. Some small comments but I'll love having this in the standard library.

src/glang/lib/std/wrappers.c3
antroseco: We also have this in the immutable versions
```
function [T] make<T&> : (value : T&) -> Optional<T&> = {
    return { ref_to_addr(&value) };
}
```

I'll change this too then

PR #128, opened 2023-09-24
Allow automatic initializer list type deduction

### LINKED ISSUE #124: Type deduction for partially specialized functions is overly pessimistic
```
typedef A : {}
typedef B : {}

function make<A> : (val : A) -> int = { return {};}

function [@Len] make<B> : (buffer : u8[@Len]&) -> B = {
    return {};
}

function main : () -> int = {
    let a : A = make<A>({});
    return 0;
}

```

Gives the error `Error: cannot deduce generic type '@Len' in function 'make<B, @Len> : (u8[@Len]&) -> B', manual specialization is required` even though the first specialization argument should eliminate `make<B>` as a possibility.

DaveDuck321:
Initializer lists have no `underlying_type` despite being a `TypedExpression`.  This made the whole interface kind of confusing to use. Now `TypedExpression` no longer requires an underlying type, instead, it has a `result_type` (which is the type an expression would collapse to if no implicit conversion was executed).

This fixes #124 and a few other confusing template/ initializer list interactions.

These fixes allow us to *finally* match the type behavior described in the docs (namely initializer lists creating new "compatible" structs).

These fixes also exposed the awkward behavior where implicit conversion is disabled when using specialized templates so I had to fix this for the parser to compile.

```
function [T] fn : (a : T, b : T) -> ...
fn<isize>(8, 8); // 8 is now implicitly converted to isize due to the explicit specialization
fn(8, isize(8)) // The function fails to specialize since this would require an implicit conversion.
```

As a bonus, this makes something like `&any` very easy to implement... It would just be a `TypedExpression` that collapses into the best reference type available but can also be implicitly converted to any reference type.

### DIFF
--- a/docs/types_overview.md
+++ b/docs/types_overview.md
@@ -5,8 +5,8 @@
 The syntax for aliasing types is:
 
 ```c
-typedef Name1 : TypeThatWeWantToAlias;
-typedef Name2 : TypeThatWeWantToAlias;
+typedef Name1 : TypeThatWeWantToAlias
+typedef Name2 : TypeThatWeWantToAlias
 ```
 `Name1`, `Name2`, and `TypeThatWeWantToAlias` are equivalent and can be used
 interchangeably in any context.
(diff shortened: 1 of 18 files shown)


antroseco:
Guh, sorry I missed this. Looks really good though

PR #129, opened 2023-10-07
Add floating point numbers

DaveDuck321:


src/glang/codegen/builtin_callables.py
```diff
@@ -99,13 +101,17 @@ def __init__(
...
         assert len(specialization) == 0
         lhs_definition = lhs.underlying_type.definition
         rhs_definition = rhs.underlying_type.definition
-        assert isinstance(lhs_definition, (IntegerDefinition, BoolDefinition))
-        assert isinstance(rhs_definition, (IntegerDefinition, BoolDefinition))
+        assert isinstance(
+            lhs_definition, (IntegerDefinition, BoolDefinition, IEEEFloatDefinition)
+        )
+        assert isinstance(
+            rhs_definition, (IntegerDefinition, BoolDefinition, IEEEFloatDefinition)
```
antroseco, line 122: The tuple is kinda verbose, should we make it a type alias?

src/glang/codegen/builtin_callables.py
```diff
@@ -138,6 +146,8 @@ def generate_ir(self, reg_gen: Iterator[int]) -> list[str]:
             assert isinstance(self._arg_type.definition, BoolDefinition)
             ir = self.UNSIGNED_IR
 
+        assert ir is not None
```
antroseco, line 158: Would it be cleaner if we defined these
```py
    FLOATING_POINT_IR = None
    SIGNED_IR = None
    UNSIGNED_IR = None
    USER_FACING_NAME = None
```
as abstract properties, so that we don't have to check if it's `None` or not

src/glang/codegen/builtin_callables.py
```diff
@@ -225,25 +240,28 @@ class BitwiseXorExpression(ArithmeticExpression):
...
 
-class CompareExpression(BasicIntegerExpression):
+class CompareExpression(BasicNumericExpression):
     @staticmethod
     def get_result_type(arguments: list[TypedExpression]) -> Type:
         return BoolType()
 
 
 class IsEqualExpression(CompareExpression):
+    FLOATING_POINT_IR = "fcmp oeq"  # Ordered and equal (can we relax this?)
```
antroseco, line 259: Do the ordered and unordered versions produce different assembly? Unordered is presumably faster, although ordered might produce slightly saner code?

Not even sure how a QNAN would come about though

src/glang/codegen/builtin_types.py
```diff
@@ -329,6 +329,28 @@ def __init__(self) -> None:
         super().__init__("iptr", get_ptr_type_info().size.in_bits, True)
 
 
+class IEEEFloatDefinition(PrimitiveDefinition):
+    # NOTE: we use IEEE rather than the target specific SIMD types
```
antroseco, line 333: Does the front end need to be target aware then when compiling floating point code?

src/glang/codegen/type_conversions.py
```diff
@@ -192,7 +201,16 @@ def last_type() -> Type:
         and last_def.bits < dest_def.bits
     ):
         promotion_cost += dest_def.bits // last_def.bits
-        expr_list.append(PromoteInteger(last_expr(), dest_type))
+        expr_list.append(PromoteNumeric(last_expr(), dest_type))
```
antroseco, line 204: Should we use `.size` here too for consistency?

src/glang/parser/parser.c3
```diff
@@ -267,6 +282,38 @@ function get_next_decimal_token : (cursor : Cursor mut&) -> ProgramSlice = {
         if ascii_is_digit(next_char) {
             cursor.position += 1;
             token.end_offset += 1;
+
+            if cursor:get_char() == char(".") {
```
antroseco, line 286: This looks a bit weird. Is their a reason you've nested these inside `if ascii_is_digit(next_char)`? I suppose we don't have `else if` yet and you don't need to check if we've seen a digit already. 

What happens if `cursor:is_end_of_token() == true` and we call `cursor:get_char()` anyway?

src/glang/parser/parser.c3
```diff
@@ -959,10 +1000,11 @@ function parse_maybe_unary_or_borrow_expression : (cursor : Cursor mut&) -> Expr
 
     if cursor:try_skip_next("&") {
         let is_mut : bool = cursor:try_skip_next_and_surrounding_white_space("mut");
+        let is_any : bool = !is_mut and cursor:try_skip_next_and_surrounding_white_space("any");
         let rhs : Expression = cursor:parse_maybe_unary_or_borrow_expression();
 
         let borrow : Borrow mut& = &mut cursor.allocator:allocate<Borrow>();
-        borrow = {is_mut, rhs, meta(start_offset, borrow.meta.end_offset)};
+        borrow = {is_mut, is_any, rhs, meta(start_offset, borrow.meta.end_offset)};
         return {11, ref_to_addr(&borrow), borrow.meta};
```
antroseco, line 1008: Sneaky!

src/glang/codegen/floats.py
```diff
@@ -0,0 +1,124 @@
...
+    # We support both exponential and decimal notation
+    # First simplify the algorithm by inserting the implicit exponent
+    if "e" not in float_str:
+        float_str += "e0"
+
+    # Now trim the sign (if any)
+    # IEEE floats have symmetric precision so both cases are identical
+    is_negative = float_str[0] == "-"
+    if float_str[0] == "-" or float_str[0] == "+":
+        float_str = float_str[1:]
```
antroseco, line 39: `float_str = float_str.lstrip("+-")`

src/glang/codegen/floats.py
```diff
@@ -0,0 +1,124 @@
...
+    if "." in mantissa_str_10:
+        decimal_point_index = mantissa_str_10.index(".")
+
+        mantissa_str_10 = mantissa_str_10.replace(".", "")
+        exponent_10 += decimal_point_index - len(mantissa_str_10)
+
+    mantissa_10 = int(mantissa_str_10)
+
+    if mantissa_10 == 0:
+        # Note: I'm discarding '-0.0' here... I'm not sure if its worth propagating
```
antroseco, line 56: Do we parse the leading `-` in integer literals?

src/glang/codegen/floats.py
```diff
@@ -0,0 +1,124 @@
...
+        # max(...) is due to LLVM weirdness
+        return "0x" + "00" * (8 * max(float_size_in_bits, 64))
+
+    # I should probably look at an official algorithm here...
+    # Its a fun problem tho, I'll just derive something
+    #  See if you can work out where my magic numbers came from :-D
```
antroseco, line 62: One day!

tests/floating_point/basic_use.c3
```diff
@@ -0,0 +1,19 @@
...
+    // Check that we can parse everything
+    let e : f32 = 1.0;
+    let f : f32 = 1e30;
+    let g : f32 = 1e+9;
+    let h : f32 = 0.0;
+    return 0;
+}
+
+/// @COMPILE
+/// @RUN
```
antroseco, line 19: `@GREP_IR` might be useful here if you want to test your algorithm

antroseco (commented):
Very clean! I haven't studied the float-to-hex algorithm yet though

src/glang/codegen/builtin_callables.py
DaveDuck321: Hmm, ur right. Sometimes it doesn't make sense to define them all eg. bitwise or doesn't need a floating point version... But this seems like something oop can fix

src/glang/codegen/builtin_types.py
DaveDuck321: Hmm, probably yes... Good point.... I think I need to add something to our `targets/*.yml`

src/glang/parser/parser.c3
DaveDuck321: Ahhh. Thats a secret. Damn my leaky diff.

DaveDuck321: Probably whats breaking the tests too :upside_down_face:

src/glang/codegen/floats.py
DaveDuck321: Sometimes? We do in generics but not in expressions. Floats should behave in exactly the same way but since they're not allowed in generics ur right, `-0.0` won't actually be parsable even if I implemented it.

src/glang/codegen/builtin_callables.py
antroseco: Maybe OOP would be nicer, although I'd rather avoid it if possible

src/glang/codegen/builtin_types.py
antroseco: That would do, although we could just hold off until we actually need to introduce a distinct implementation (a quick search indicates the IEEE representation is very widely used)

src/glang/codegen/floats.py
antroseco: Guhh, that's not very nice. Compile-time expressions would make this more consistent :)

src/glang/codegen/builtin_callables.py
DaveDuck321: I've looked at some x86 assembly and you're right, the unordered versions produce faster code. I've spent some time thinking, and I think it is fine to stop requiring the order: we're already ignoring integer overflows so why not also ignore NaNs.

src/glang/parser/parser.c3
DaveDuck321: Yeah, I'm not really a fan of this tbh. We'd need better control flow to improve this ergonomically tho (`if` `else` at a minimum and ideally `break`/`continue`).

> What happens if cursor:is_end_of_token() == true and we call cursor:get_char() anyway?

This is fine, get_char() will just return whatever character caused the token to end (which here I'm fairly should be the "."). If `cursor:is_end_of_token() == true` and the next character is not the dot, we skip both branches and the loop ends normally.

PR #131, opened 2023-11-19
Run tests with LLVM 17

antroseco:
Can't use Python 3.12 yet unfortunately (not easily at least)

### DIFF
--- a/.github/workflows/graphene_tests.yml
+++ b/.github/workflows/graphene_tests.yml
@@ -9,7 +9,7 @@ jobs:
   test:
     strategy:
       matrix:
-        llvm-version: [15, 16]
+        llvm-version: [16, 17]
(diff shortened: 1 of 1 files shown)


.github/workflows/graphene_tests.yml
```diff
@@ -9,7 +9,7 @@ jobs:
   test:
     strategy:
       matrix:
-        llvm-version: [15, 16]
+        llvm-version: [16, 17]
```
DaveDuck321, line 12: Yeah, no reason to support LLVM 15 anymore, Manjaro's on 16.

DaveDuck321 (approved):

.github/workflows/graphene_tests.yml
antroseco: Mhm, I think last two is pretty reasonable

PR #135, opened 2023-12-16
Small QoL improvements

antroseco:
- Function arguments that were passed by value should be mutable
- Handle arguments passed to `hatch run test:run`
- Allow UFCS calls to match with both `mut&` and `&` (`mut&` preferred)
- Fix debug configuration

### DIFF
--- a/src/glang/codegen/builtin_types.py
+++ b/src/glang/codegen/builtin_types.py
@@ -10,7 +10,6 @@ from .interfaces import SpecializationItem, Type, TypeDefinition, format_special
 from .user_facing_errors import (
     ArrayDimensionError,
     FailedLookupError,
-    GrapheneError,
     InvalidIntSize,
     VoidArrayDeclaration,
     VoidStructDeclaration,
@@ -26,11 +25,8 @@ class PlaceholderDefinition(TypeDefinition):
         assert False
 
     def copy_with_storage_kind(self, parent: Type, kind: Type.Kind) -> Type:
-        if kind.is_reference():
-            return AnonymousType(
-                ReferenceDefinition(parent, kind == Type.Kind.MUTABLE_REF)
-            )
-        assert False
+        assert kind.is_reference()
+        return AnonymousType(ReferenceDefinition(parent, kind))
 
     @property
     def is_finite(self) -> bool:
@@ -159,13 +155,14 @@ class ValueTypeDefinition(TypeDefinition):
     def copy_with_storage_kind(self, parent: Type, kind: Type.Kind) -> Type:
         assert self.storage_kind != kind
         assert kind.is_reference()
-        return AnonymousType(ReferenceDefinition(parent, kind == Type.Kind.MUTABLE_REF))
+        return AnonymousType(ReferenceDefinition(parent, kind))
 
 
 class PtrTypeDefinition(TypeDefinition):
-    def __init__(self, is_mut: bool) -> None:
+    def __init__(self, storage_kind: Type.Kind) -> None:
         super().__init__()
-        self.is_mut = is_mut
+        # Illegal value types are allowed (used by arrays).
+        self._storage_kind = storage_kind
 
     @abstractmethod
     def copy_with_storage_kind(self, _: Type, kind: Type.Kind) -> Type:
@@ -189,9 +186,11 @@ class PtrTypeDefinition(TypeDefinition):
 
     @property
     def storage_kind(self) -> Type.Kind:
-        if self.is_mut:
-            return Type.Kind.MUTABLE_REF
-        return Type.Kind.CONST_REF
+        return self._storage_kind
+
+    @property
+    def is_mut(self) -> bool:
+        return self.storage_kind.is_mutable_reference()
 
 
 class PrimitiveDefinition(ValueTypeDefinition):
@@ -461,8 +460,9 @@ class StackArrayDefinition(ValueTypeDefinition):
 
 
 class ReferenceDefinition(PtrTypeDefinition):
-    def __init__(self, value_type: Type, is_mut: bool) -> None:
-        super().__init__(is_mut)
+    def __init__(self, value_type: Type, storage_kind: Type.Kind) -> None:
+        assert storage_kind.is_reference()
+        super().__init__(storage_kind)
 
         # TODO: user-facing error
         assert not value_type.storage_kind.is_reference()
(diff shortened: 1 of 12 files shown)


DaveDuck321 (approved):
Cool, looks good.
The immutable function parameters was intentional originally but no use keeping it around if it's annoying 😇.

antroseco:
> Cool, looks good. The immutable function parameters was intentional originally but no use keeping it around if it's annoying 😇.

Do you remember why we did that?

I also fixed the annoying UFCS case where you want to call a function that takes a `mut` this as a `&`. That was a bit more involved than the other changes...

tests/constants/ufcs.c3
```diff
@@ -1,21 +1,24 @@
...
+    return 0;
 }
 
 function main : () -> int = {
-    mut a : A = {1};
-    return a:run();
+    mut a : A = {};
+    // Can match with A&, but should prefer A mut&.
+    return a:foo() + a:bar();
 }
```
DaveDuck321, line 21: Cool. Much more intuitive

src/glang/graphene_parser.py
```diff
@@ -653,7 +666,12 @@ def Borrow(self, node: lp.Borrow) -> FlattenedExpression:
         # TODO: make this const-correct
         # I've temporarily disabled this so we can parse the parser as it
         #  transitions to using the new syntax
```
DaveDuck321, line 668: This comment is likely outdated

src/glang/graphene_parser.py
```diff
@@ -585,19 +587,30 @@ def _function_call_outer(
...
+                    # (&a):fn(), (&mut a):fn(), or rvalue():fn();
+                    # We do not allow the 3rd option
+                    possible = this_arg.expression().result_type_as_if_borrowed
+                    if possible.storage_kind == cg.Type.Kind.VALUE:
+                        raise BorrowWithNoAddressError(
+                            possible.format_for_output_to_user()
+                        )
+                case _:
+                    # cg.Type.Kind.MUTABLE_OR_CONST_REF shouldn't be possible.
+                    raise AssertionError()
```
DaveDuck321, line 613: Ooo, very modern python here.

src/glang/codegen/type_conversions.py
```diff
@@ -152,21 +152,31 @@ def last_type() -> Type:
...
-            src.format_for_output_to_user(),
-            dest_type.format_for_output_to_user(True),
-            maybe_missing_borrow,
-        )
+            raise TypeCheckerError(
+                context,
+                src.format_for_output_to_user(),
+                dest_type.format_for_output_to_user(True),
+                maybe_missing_borrow,
+            )
```
DaveDuck321, line 179: Hmmm, fine for now but I'm not convinced this is more readable than an if-else chain :thinking:. Maybe it'll be fine when I get more used to the syntax

src/glang/codegen/type_conversions.py
```diff
@@ -152,21 +152,31 @@ def last_type() -> Type:
...
+        case Type.Kind.MUTABLE_OR_CONST_REF, dest_kind if dest_kind.is_reference():
+            expr_list.append(
+                Reinterpret(
+                    last_expr(),
+                    last_type().convert_to_storage_type(dest_kind),
+                )
+            )
+            # Preferentially select the mutable overload.
+            if dest_kind == Type.Kind.CONST_REF:
+                promotion_cost += 1
```
DaveDuck321, line 165: Clean

src/glang/codegen/translation_unit.py
```diff
@@ -68,7 +68,13 @@ def _add_parameter(self, param_name: str, param_type: Type):
...
         self.top_level_scope.add_generatable(fn_param)
 
-        fn_param_var = StackVariable(param_name, param_type, False, True)
+        # Allow assignments to parameters passed by value (but our references
+        # still need to be immutable).
+        is_value_type = param_type.storage_kind == Type.Kind.VALUE
+
+        fn_param_var = StackVariable(
+            param_name, param_type, is_mutable=is_value_type, initialized=True
+        )
```
DaveDuck321, line 77: Fair. This breaks the const-by-default assumption but it also avoids redundant copies when you actually need to mutate.

src/glang/codegen/interfaces.py
```diff
@@ -64,10 +64,15 @@ class Kind(Enum):
         VALUE = 1
         MUTABLE_REF = 2
         CONST_REF = 3
+        # Mutable reference, but implicitly convertible to a const reference.
+        MUTABLE_OR_CONST_REF = 4
```
DaveDuck321, line 68: This *sometimes valid* enum seems like it'll increase cyclomatic complexity... but this commit is very clean regardless, this seems like a good pragmatic approach.

src/glang/codegen/builtin_types.py
```diff
@@ -472,8 +473,15 @@ def copy_with_storage_kind(self, parent: Type, kind: Type.Kind) -> Type:
...
-        return parent
+        if kind == self.storage_kind:
+            return parent
+
+        assert self.storage_kind == Type.Kind.MUTABLE_OR_CONST_REF
+
+        # TODO there must be a better way.
+        new_definition = copy(self)
+        new_definition._storage_kind = kind
+        return AnonymousType(new_definition)
```
DaveDuck321, line 484: This is a little scary... This method made our type resolution kinda buggy by copying classes we were referencing, I fixed it by removing all the copies so we could use the `UnresolvedReference` class.... If this causes issues I suppose we're gonna have to rethink how either of these things work.

DaveDuck321 (approved):
Very nice. Good to merge as-is (preferably keeping the commit structure).
After this is merged we can remove some of the duplicated member functions in the standard library.

src/glang/graphene_parser.py
antroseco: Removed

src/glang/codegen/type_conversions.py
antroseco: It's new and cool so it must be better ;)

I just wanted an excuse to use pattern matching lol

src/glang/graphene_parser.py
antroseco: Here I think it's actually more readable

src/glang/codegen/builtin_types.py
antroseco: > by copying classes we were referencing

Hmm, not sure I understand why that's problematic. I've changed it to return `AnonymousType(ReferenceDefinition(...))` to avoid the copy. If you still don't like it then open an issue pls

src/glang/codegen/interfaces.py
antroseco: Well... it could always be valid, we just don't have explicit borrow syntax for it :)

antroseco: But yeah, I think this is easier to implement on top of what we have, compared to doing two lookups

src/glang/codegen/translation_unit.py
antroseco: I do like the const-by-default, but we don't really have syntax for mutable arguments. Maybe we should

PR #136, opened 2023-12-17
Emit basic debugging information

antroseco:
I think you can take a look now - it's pretty functional.

The compiler now always emits LLVM-style debugging information (basically DWARF).
- The format is pretty straightforward, but it might help to look at some example IR before you review
- To get line number support, we need to pass the `Meta` from the parser to every single expression we codegen
- That's quite tedious unfortunately, but there's no other way
- The `generate_ir()` function now needs more inputs and has a second output, to handle the metadata
- To get type and variable support (`gdb` needs to know how to read variables off of memory), we need to serialize each type
- That's quite tricky and I don't really have a nice solution yet - maybe we should do it during type resolution to avoid cycles, etc but also this is strictly an IR thing so it doesn't feel right either

### DIFF
--- a/src/glang/codegen/builtin_callables.py
+++ b/src/glang/codegen/builtin_callables.py
@@ -1,6 +1,7 @@
 from abc import abstractmethod
 from typing import Iterator
 
+from ..parser.lexer_parser import Meta
 from .builtin_types import (
     BoolDefinition,
     BoolType,
@@ -18,7 +19,10 @@ from .user_facing_errors import OperandError
 class BuiltinCallable(StaticTypedExpression):
     @abstractmethod
     def __init__(
-        self, specialization: list[SpecializationItem], arguments: list[TypedExpression]
+        self,
+        specialization: list[SpecializationItem],
+        arguments: list[TypedExpression],
+        meta: Meta,
     ) -> None:
         pass
 
@@ -28,7 +32,10 @@ class UnaryIntegerExpression(BuiltinCallable):
     USER_FACING_NAME = ""
 
     def __init__(
-        self, specialization: list[SpecializationItem], arguments: list[TypedExpression]
+        self,
+        specialization: list[SpecializationItem],
+        arguments: list[TypedExpression],
+        meta: Meta,
     ) -> None:
         (self._arg,) = arguments
 
@@ -39,7 +46,7 @@ class UnaryIntegerExpression(BuiltinCallable):
         assert isinstance(definition, (IntegerDefinition, BoolDefinition))
 
         self._arg_type = self._arg.result_type.convert_to_value_type()
-        StaticTypedExpression.__init__(self, self._arg_type, Type.Kind.VALUE)
+        StaticTypedExpression.__init__(self, self._arg_type, Type.Kind.VALUE, meta)
 
     def __repr__(self) -> str:
         return f"{self.__class__.__name__}({self._arg})"
@@ -95,7 +102,10 @@ class BasicIntegerExpression(BuiltinCallable):
         pass
 
     def __init__(
-        self, specialization: list[SpecializationItem], arguments: list[TypedExpression]
+        self,
+        specialization: list[SpecializationItem],
+        arguments: list[TypedExpression],
+        meta: Meta,
     ) -> None:
         lhs, rhs = arguments
         # This is not a user-facing function, we don't need sensible error messages
@@ -109,7 +119,7 @@ class BasicIntegerExpression(BuiltinCallable):
         assert lhs.result_type == rhs.result_type
 
         StaticTypedExpression.__init__(
-            self, self.get_result_type(arguments), Type.Kind.VALUE
+            self, self.get_result_type(arguments), Type.Kind.VALUE, meta
         )
 
         self._arg_type = lhs.result_type.convert_to_value_type()
@@ -256,6 +266,7 @@ class AlignOfExpression(BuiltinCallable):
         self,
         specialization: list[SpecializationItem],
         arguments: list[TypedExpression],
+        meta: Meta,
     ) -> None:
         assert len(arguments) == 0
         assert len(specialization) == 1
@@ -267,7 +278,7 @@ class AlignOfExpression(BuiltinCallable):
         )
 
         StaticTypedExpression.__init__(
-            self, self._result.underlying_type, Type.Kind.VALUE
+            self, self._result.underlying_type, Type.Kind.VALUE, meta
         )
 
     def __repr__(self) -> str:
@@ -292,6 +303,7 @@ class SizeOfExpression(BuiltinCallable):
         self,
         specialization: list[SpecializationItem],
         arguments: list[TypedExpression],
+        meta: Meta,
     ) -> None:
         assert len(arguments) == 0
         assert len(specialization) == 1
@@ -301,7 +313,7 @@ class SizeOfExpression(BuiltinCallable):
         self._result = ConstantExpression(SizeType(), str(self._argument_type.size))
 
         StaticTypedExpression.__init__(
-            self, self._result.underlying_type, Type.Kind.VALUE
+            self, self._result.underlying_type, Type.Kind.VALUE, meta
         )
 
     def __repr__(self) -> str:
@@ -326,6 +338,7 @@ class NarrowExpression(BuiltinCallable):
         self,
         specialization: list[SpecializationItem],
         arguments: list[TypedExpression],
+        meta: Meta,
     ) -> None:
         (self._argument,) = arguments
         (return_type,) = specialization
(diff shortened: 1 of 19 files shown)


src/glang/codegen/debug.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+@dataclass
+class DICompileUnit(Metadata):
+    file: DIFile
+
+    def __repr__(self) -> str:
+        # TODO print glang version.
+        # NOTE clang checks that the language is valid... so let's pretend we
+        # are C.
```
DaveDuck321, line 37: Lol :cowboy_hat_face:

src/glang/codegen/translation_unit.py
```diff
@@ -151,23 +195,29 @@ def lookup_call_expression(
...
         encoded_str, encoded_length = encode_string(string)
         str_type = AnonymousType(CharArrayDefinition(encoded_str, encoded_length))
         static_storage = StaticVariable(
-            f'string literal: "{string}"', str_type, False, encoded_str
+            f'string literal: "{string}"',
+            str_type,
+            False,
+            encoded_str,
+            meta,
+            self.di_files[0],  # FIXME this is wrong.
```
DaveDuck321, line 220: Seems like `Meta` should have a file path

src/glang/codegen/translation_unit.py
```diff
@@ -180,22 +230,59 @@ def add_static_variable(self, var: StaticVariable) -> None:
...
-        lines.extend(self.symbol_table.get_ir_for_initialization())
+        output.lines.append("")
+        output.lines.extend(self.symbol_table.get_ir_for_initialization())
 
-        lines.append("")
+        # We need to declare any debugger intrinsics we use for some reason.
+        output.lines.append("")
+        output.lines.append(
+            "declare void @llvm.dbg.declare(metadata, metadata, metadata)"
+        )
```
DaveDuck321, line 257: How strange

src/glang/codegen/translation_unit.py
```diff
@@ -180,22 +230,59 @@ def add_static_variable(self, var: StaticVariable) -> None:
...
+
+        output.lines.append("")
         for func in self._fn_bodies:
-            lines.extend(func.generate_ir())
+            output.extend(func.generate_ir(self._metadata_gen))
+
+        # behaviour = 1 (Error)
+        # Emits an error if two values disagree, otherwise the resulting value
+        # is that of the operands.
```
DaveDuck321, line 265: Not sure I understand this comment tbh...
Oh `MetadataFlag.behaviour = 1`... when is the error emitted, is this when the debugger attaches?

src/glang/graphene_parser.py
```diff
@@ -503,9 +515,7 @@ def UnaryOperatorUse(self, node: lp.UnaryOperatorUse) -> FlattenedExpression:
 
         flattened_expr = FlattenedExpression(rhs.subexpressions)
         call_expr = self._program.lookup_call_expression(
-            node.name,
-            [],  # Don't specialize operators
-            [rhs.expression()],
+            node.name, [], [rhs.expression()], node.meta  # Don't specialize operators
```
DaveDuck321, line 518: This comment should probably be deleted now it's moved, I don't think its very useful anymore.

tests/run_tests.py
```diff
@@ -147,7 +147,7 @@ def run_test(file_path: Path) -> bool:
 
     # @GREP_IR
     for needle in config.grep_ir_strs:
-        if needle not in ir_output:
```
DaveDuck321, line 150: Cool, nice. cleanup.

src/glang/codegen/debug.py
```diff
@@ -0,0 +1,290 @@
...
+    class_type = "DW_TAG_class_type"
+    entry_point = "DW_TAG_entry_point"
+    enumeration_type = "DW_TAG_enumeration_type"
+    formal_parameter = "DW_TAG_formal_parameter"
+    imported_declaration = "DW_TAG_imported_declaration"
+    label = "DW_TAG_label"
+    lexical_block = "DW_TAG_lexical_block"
+    member = "DW_TAG_member"
+    pointer_type = "DW_TAG_pointer_type"
+    reference_type = "DW_TAG_reference_type"
```
DaveDuck321, line 21: Vim go brrr

DaveDuck321 (approved):
I had to read through the LLVM documentation but everything makes sense, I'm surprised you managed this with such a small diff. I really like the new abstractions: `IRContext` is much better than passing generators around and everything in `debug.py` looks nice.

While this isn't *fully* complete atm it's probably good to merge as is to avoid a nasty merge, everything else can be done in separate PRs.

src/glang/codegen/builtin_callables.py
```diff
@@ -39,19 +45,20 @@ def __init__(
...
         self._arg_type = self._arg.result_type.convert_to_value_type()
-        StaticTypedExpression.__init__(self, self._arg_type, Type.Kind.VALUE)
+        StaticTypedExpression.__init__(self, self._arg_type, Type.Kind.VALUE, meta)
 
     def __repr__(self) -> str:
         return f"{self.__class__.__name__}({self._arg})"
 
-    def generate_ir(self, reg_gen: Iterator[int]) -> list[str]:
+    def generate_ir(self, ctx: IRContext) -> IROutput:
+        # https://llvm.org/docs/LangRef.html#add-instruction (and below)
```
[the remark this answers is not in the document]
antroseco: Woops, thanks

src/glang/codegen/translation_unit.py
```diff
@@ -64,20 +80,22 @@ def __init__(
                 )
 
     def _add_parameter(self, param_name: str, param_type: Type):
-        fn_param = FunctionParameter(param_type)
+        # FIXME pass the true meta.
```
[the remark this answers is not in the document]
antroseco: Yeah, so the meta comes from the function definition (not sure if that's the token corresponding to the name or the `function`/`foreign` keywords). It's good enough for now, but as you said,  we can have parameters spanning multiple lines

src/glang/codegen/translation_unit.py
```diff
@@ -64,20 +80,22 @@ def __init__(
...
         self.top_level_scope.add_generatable(fn_param)
 
         # Allow assignments to parameters passed by value (but our references
         # still need to be immutable).
         is_value_type = param_type.storage_kind == Type.Kind.VALUE
 
+        # FIXME pass the true meta.
         fn_param_var = StackVariable(
-            param_name, param_type, is_mutable=is_value_type, initialized=True
+            param_name, param_type, is_value_type, True, self._meta, self._di_file
```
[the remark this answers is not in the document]
antroseco: This is slightly incorrect, which IMO is better than nothing: gdb still takes you to almost the right place.

We can't use poison in metadata, but you can usually leave it empty. It just means you get no source information.

src/glang/codegen/translation_unit.py
```diff
@@ -64,20 +80,22 @@ def __init__(
...
         )
         self.top_level_scope.add_variable(fn_param_var)
 
-        fn_param_var_assignment = VariableInitialize(fn_param_var, fn_param)
+        fn_param_var_assignment = VariableInitialize(fn_param_var, fn_param, self._meta)
```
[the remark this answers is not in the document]
antroseco: Added

src/glang/codegen/translation_unit.py
antroseco: That would be a really nice refactor

src/glang/codegen/translation_unit.py
antroseco: Sorry I was being lazy. I made it an enum now.

Basically, this behavior flag determines how LLVM deals with conflicting metadata when merging two or more "modules" together.

Error means that compilation/linking should fail.

src/glang/graphene_parser.py
antroseco: Oops, I just missed the trailing comma and black re-arranged everything. I fixed it now.

src/glang/codegen/debug.py
antroseco: Yes haha

antroseco:
> I had to read through the LLVM documentation but everything makes sense, I'm surprised you managed this with such a small diff. I really like the new abstractions: `IRContext` is much better than passing generators around and everything in `debug.py` looks nice.

Yeah, I think `IRContext` is really nice, and we could use it for other things in the future (e.g. passing in the name of the last label).

> While this isn't _fully_ complete atm it's probably good to merge as is to avoid a nasty merge, everything else can be done in separate PRs.

Still a lot of work to do - this is just the bare minimum we need for gdb to be useful!

src/glang/codegen/translation_unit.py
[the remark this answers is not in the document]
DaveDuck321: Make sense, this doesn't seem very confusing. I'm completely ok with this.

PR #137, opened 2023-12-28
Add floating point stage to bootstrap

DaveDuck321:
In #129, I wanted to add a floating point `abs` to the standard library but this would break the bootstrap since it required the literal `0.0`.

This PR fixes the bootstrap.

### DIFF
--- a/glang.spec
+++ b/glang.spec
@@ -1,5 +1,5 @@
 Name:           glang
-Version:        0.2.1
+Version:        0.3.0
 Release:        %autorelease
 Summary:        A Graphene front-end for LLVM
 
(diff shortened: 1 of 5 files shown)


glang.spec
```diff
@@ -1,5 +1,5 @@
 Name:           glang
-Version:        0.2.1
+Version:        0.3.0
```
antroseco, line 2: Can you update this version as well please
https://github.com/DaveDuck321/GrapheneLang/blob/da3b7d5e7c8289045bd2beb8ea425daf5aa098f4/pyproject.toml#L8

One day we'll have a single version number...

antroseco (approved):

glang.spec
DaveDuck321: Thanks :blush:. I knew I was forgetting something

PR #139, opened 2023-12-30
Implement Itanium C++ ABI name mangling

antroseco:
Based on https://itanium-cxx-abi.github.io/cxx-abi/abi.html#mangling, which is what GCC and Clang use. 

Everything works (by inspection), but I'll add tests before merging.

I haven't implemented compression yet (and not sure it's worth it), but it's necessary if we want to be 100% compatible with other toolchains.

### DIFF
--- a/src/glang/codegen/builtin_types.py
+++ b/src/glang/codegen/builtin_types.py
@@ -19,6 +19,13 @@ from .debug import (
 )
 from .floats import float_literal_to_exact_hex
 from .interfaces import SpecializationItem, Type, TypeDefinition, format_specialization
+from .mangling import (
+    mangle_float,
+    mangle_int,
+    mangle_operator_name,
+    mangle_source_name,
+    mangle_template_args,
+)
 from .user_facing_errors import (
     ArrayDimensionError,
     FailedLookupError,
@@ -122,22 +129,15 @@ class NamedType(Type):
             assert self.alias is not None
             return self.alias.ir_mangle
 
-        prefix = "__T_"
-        if len(self.specialization) == 0:
-            return prefix + self.name
+        # Perhaps built-in types shouldn't be a NamedType.
+        if isinstance(self.definition, PrimitiveDefinition):
+            return self.definition.ir_mangle
 
-        specializations = []
-        for argument in self.specialization:
-            if isinstance(argument, int):
-                # TODO: negative numbers
-                specializations.append(str(argument))
-            elif isinstance(argument, Type):
-                specializations.append(argument.ir_mangle)
-            else:
-                assert False
+        source_name = mangle_source_name(self.name)
+        if not self.specialization:
+            return source_name
 
-        specialization_format = "".join(specializations)
-        return f"{prefix}{self.name}__S{specialization_format}"
+        return source_name + mangle_template_args(self.specialization)
 
     @property
     def ir_type(self) -> str:
@@ -242,10 +242,6 @@ class PrimitiveDefinition(ValueTypeDefinition):
         return self._ir
 
     @property
-    def ir_mangle(self) -> str:
-        return self._ir
-
-    @property
     def size(self) -> int:
         return self._size
 
@@ -287,6 +283,11 @@ class VoidDefinition(PrimitiveDefinition):
 
     def to_di_type(self, metadata_gen: Iterator[int]) -> list[Metadata]:
         assert False
+
+    @property
+    def ir_mangle(self) -> str:
+        # <builtin-type> ::= v  # void
+        return "v"
 
 
 class VoidType(PrimitiveType):
@@ -331,6 +332,10 @@ class IntegerDefinition(PrimitiveDefinition):
                 TypeKind.signed if self.is_signed else TypeKind.unsigned,
             )
         ]
+
+    @property
+    def ir_mangle(self) -> str:
+        return mangle_int(self.bits, self.is_signed)
 
 
 class GenericIntType(PrimitiveType):
@@ -393,6 +398,10 @@ class IEEEFloatDefinition(PrimitiveDefinition):
             )
         ]
 
+    @property
+    def ir_mangle(self) -> str:
+        return mangle_float(self.bits)
+
 
 class IEEEFloat(PrimitiveType):
     def __init__(self, size_in_bits: int) -> None:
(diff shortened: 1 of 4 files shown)


src/glang/codegen/mangling.py
```diff
@@ -0,0 +1,190 @@
...
+            return "nt"
+        case "[]":
+            return "ix"
+        # Vendors who define builtin extended operators (e.g. __imag) shall
+        # encode them as a 'v' prefix followed by the operand count as a single
+        # decimal digit, and the name in <length,ID> form.
+        case "**":
+            return "v2" + mangle_source_name("power")
+        case "@":
+            return "v2" + mangle_source_name("matmul")
```
antroseco, line 188: These show up as `operator power ` and `operator matmul`

src/glang/codegen/builtin_types.py
```diff
@@ -122,30 +129,23 @@ def ir_mangle(self) -> str:
             assert self.alias is not None
             return self.alias.ir_mangle
 
-        prefix = "__T_"
-        if len(self.specialization) == 0:
-            return prefix + self.name
+        # Perhaps built-in types shouldn't be a NamedType.
+        if isinstance(self.definition, PrimitiveDefinition):
+            return self.definition.ir_mangle
```
antroseco, line 134: On the other hand, we could output our names (e.g. `i32`), but then it becomes harder to call Graphene code from C/C++ land.

src/glang/codegen/builtin_types.py
```diff
@@ -472,8 +486,15 @@ def ir_type(self) -> str:
...
-        return f"__S{member_mangle}"
+        member_types = (m_type for _, m_type in self.members)
+        # <builtin-type> ::= u <source-name> [<template-args>]  # vendor extended type
+        # FIXME binutils rejects the above :(.
+        # `u <source-name>` works and so does `<source-name> <template-args>`,
+        # so fall back to the latter until we can fix it (seems like a quick
+        # fix in gcc/libiberty/cp-demangle.c).
+        return (
+            f"{mangle_source_name('AnonymousType')}{mangle_template_args(member_types)}"
+        )
```
DaveDuck321, line 497: Damn, good documenting. How about the llvm tooling?

src/glang/codegen/builtin_types.py
```diff
@@ -710,9 +735,10 @@ def format_for_output_to_user(self) -> str:
     @property
     def ir_mangle(self) -> str:
         assert not self.is_illegal_value_type
-        dimensions = "_".join(map(str, self.known_dimensions))
-        const = "" if self.is_mut else "C"
-        return f"__UA{const}_{self.member.ir_mangle}__{dimensions}__"
+        #  <array-type> ::= A [<array bound number>] _ <element type>
+        known_dimensions = str.join("", map(lambda d: f"A{d}_", self.known_dimensions))
+        const = "" if self.is_mut else "K"
+        return const + "PA_" + known_dimensions + self.member.ir_mangle
```
DaveDuck321, line 741: Why is this `PA_` not just `A_`?
https://itanium-cxx-abi.github.io/cxx-abi/abi.html#mangle.array-type

DaveDuck321 (approved):
I like this, I think you've made the mangling simpler overall despite the crazy Itanium spec.

It'd be cool to have a test linking against c++ with specialized template and overloading :upside_down_face: .

src/glang/codegen/builtin_types.py
DaveDuck321: I'm guessing c++ just encodes it as a pointer + fixed sized arrays?

src/glang/codegen/mangling.py
```diff
@@ -0,0 +1,189 @@
...
+    if isinstance(arg, Type):
+        # <template-arg> ::= <type>  # type or template
+        return arg.ir_mangle
+    else:
+        assert isinstance(arg, int)
+        # <template-arg> ::= <expr-primary>             # simple expressions
+        # <expr-primary> ::= L <type> <value number> E  # integer literal
+        # FIXME mangle as an int, even though we don't actually force a type.
+        # TODO negative numbers.
+        return f"Li{arg}E"
```
DaveDuck321, line 80: Wow, this spec is involved. Looks like negative numbers are quite easy to support tho.

src/glang/codegen/builtin_types.py
antroseco: That sounds reasonable, although it's a bit awkward to implement as the types would have to be defined as structs - if it's a `typedef` or `using` then it ends up mangling it as the built-in type, instead of using the name you gave it: https://godbolt.org/z/1x3qaKj1o.

Anyway, we can change this later if we ever get around to writing C++ bindings :)

src/glang/codegen/builtin_types.py
antroseco: Yeah, heap arrays decay into pointers (carried over from C no doubt)

src/glang/codegen/mangling.py
antroseco: Yep, they thought of everything :)

I'll get around to adding negative numbers soon, I just didn't want to do it in this PR

src/glang/codegen/builtin_types.py
antroseco:
```
$ llvm-cxxfilt _Z3foou1TIiiE
_Z3foou1TIiiE
```

Nope :(

Surprisingly, as this was an amendment to the spec proposed by a clang engineer!

src/glang/codegen/builtin_types.py
DaveDuck321: I think this would be fun to test. I'm sure that mangling into A_i in Graphene and calling from c++ would be fine due to ODR, I'm not convinced we'd be smart enough to link to the correct c++ function from Graphene. Maybe this is all handled in the linker and it doesn't matter.

If it makes no practical difference I'd much prefer A_i personally.

antroseco: Hmm, not sure what you mean - the linker just sees symbols and matches them verbatim. If we want to link with C++ code then we need to produce exactly matching symbols.

> If it makes no practical difference I'd much prefer A_i personally.

Happy to use `A_i` until we have a good reason to do anything else

PR #141, opened 2024-01-13
Add memory-map backed file reads

DaveDuck321:
Instead of streaming reads to a buffer, let's just avoid the extra copy and let the kernel give us the correct pages. This requires that the file can be memory mapped and `stat`ed. But the API is MUCH nicer + and no dynamic allocation is required to read the entire file, the kernel just gives us whatever pages we want -- no need for an inner loop.

I ran a test program that counted digits in a 1GB file (inside a loop), this saw a 2x speedup (I'm not sure why usr time is higher tho).

Old:
```
Executed in    4.77 secs    fish           external
   usr time    1.08 secs  190.00 micros    1.08 secs
   sys time    3.68 secs  128.00 micros    3.68 secs
```

New:
```
Executed in    2.81 secs    fish           external
   usr time    2.34 secs  189.00 micros    2.34 secs
   sys time    0.46 secs  126.00 micros    0.46 secs
```

And when I don't loop:

Old:
```
Executed in  524.10 millis    fish           external
   usr time  114.61 millis    0.00 micros  114.61 millis
   sys time  407.90 millis  298.00 micros  407.60 millis
```

New:
```
Executed in  315.17 millis    fish           external
   usr time  263.26 millis  200.00 micros  263.06 millis
   sys time   50.95 millis  133.00 micros   50.82 millis
```


DaveDuck321:
Oh god. The stat stuff is gonna be hard on arm.

src/glang/lib/std/io.c3
```diff
@@ -5,46 +5,67 @@
...
     let bytes_read: isize = sys_read(file.fd, &mut buffer.data, buffer.length);
     return {&buffer.data, bytes_read};
 }
 
-function close : (file : UnbufferedFile&) -> void = {
+function close : ( file : File& ) -> void = {
     if file.fd >= 3 {
         // Don't close std{in,out,err}
-        runtime_assert(sys_close(file.fd) > 0);
+        runtime_assert(sys_close(file.fd) >= 0);
```
antroseco, line 29: Actually, does `close()` ever return `> 0`?

src/glang/lib/std/io.c3
```diff
@@ -53,13 +74,13 @@ function flush : (io : BufferedFile mut&) -> void = {
     let data_to_write : View<u8> = {&io.buffer, io.buffer_pointer};
     let string_to_write : StringView = {data_to_write};
     // UFCS where `write` takes a `T&` and we have a `T mut&`
-    (&io.underlying_file):write(data_to_write);
+    (&io.underlying):write(data_to_write);
```
antroseco, line 77: Can we do `io.underlying:write(data_to_write` now? (here and below)

src/glang/lib/std/io.c3
```diff
@@ -110,7 +121,7 @@ function puts : (str : StringView) -> int = {
     // Otherwise, it shall return EOF,
     // TODO: shall set an error indicator for the stream,
     // TODO: and errno shall be set to indicate the error.
-    return EOF();
```
antroseco, line 113: Why?

src/glang/lib/std/io.c3
```diff
@@ -81,16 +102,6 @@ function consume_length : (io : BufferedFile mut&, length : isize) -> void = {
     io.buffer_pointer += length;
 }
 
-function read : (file : BufferedFile&, buffer: Span<u8>) -> View<u8> = {
-    return file.underlying_file:read(buffer);
-}
```
antroseco, line 86: I guess we don't need this

src/glang/lib/std/sys/stat.c3
```diff
@@ -0,0 +1,47 @@
...
+    st_atim : timespec,
+    st_mtim : timespec,
+    st_ctim : timespec,
+    __glibc_reserved : isize[3],
+}
+
+function stat : ( fd : int ) -> Optional<stat_t> = {
+    // TODO: propagate the actual error code
+    mut stat_buf : stat_t;
+    let result : int = sys_fstat(fd, &reinterpret_addr<u8[&]>(&mut stat_buf));
```
antroseco, line 42: looks a bit weird to borrow `&mut` first and then `&`

src/glang/lib/std/sys/stat.c3
```diff
@@ -0,0 +1,47 @@
...
+    st_size : off_t,
+    st_blksize : blksize_t,
+    st_blocks : blkcnt_t,
+    st_atim : timespec,
+    st_mtim : timespec,
+    st_ctim : timespec,
+    __glibc_reserved : isize[3],
+}
+
+function stat : ( fd : int ) -> Optional<stat_t> = {
```
antroseco, line 39: Should we not call this `fstat`? It takes an `fd`, whilst POSIX `stat` takes a path.

src/glang/lib/std/sys/stat.c3
```diff
@@ -0,0 +1,47 @@
...
+@require_once "types.c3"
+
+// This is VERY difficult to get right. fstat takes a `stat_t` buffer BUT we
+// don't actually know the layout of `stat_t` because it depends on the kernel's
+// compile flags. POSIX just guarantees that the struct contains a minimum
+// subset of fields.
+
+// https://man7.org/linux/man-pages/man3/stat.3type.html
+// The below structs are generated via the following command:
+// echo "#include <sys/stat.h>" | g++ -xc -E -
```
antroseco, line 14: I feel like the only sane/sustainable approach here is to auto-generate code like this during bootstrap/package build time.

src/glang/lib/std/sys/types.c3
```diff
@@ -1,5 +1,15 @@
...
+typedef dev_t : u64
+typedef gid_t : id_t
+typedef id_t : u32
+typedef ino_t : u64
+typedef mode_t : u32
+typedef nlink_t : u64
 typedef off_t : isize
 typedef pid_t : int
+typedef time_t : isize
+typedef uid_t : id_t
```
antroseco, line 15: These are also platform-dependent and should be auto-generated.

src/glang/parser/parser.c3
```diff
@@ -1396,27 +1396,24 @@ function parse_next_top_level_feature : (cursor : Cursor mut&) -> TopLevelFeatur
...
-    mut allocator : Allocator = initialize_allocator();
-    mut file_contents : Vector<u8> = make<Vector<u8>>(&mut allocator);
-
     // 1) Read the entire file
-    {
-        mut read_buffer : u8[1024];
+    let file_to_parse : CString = {&argv[1]};
+    let file : Optional<File> = open(file_to_parse, O_RDONLY());
+    if !file:has_value() {
+        printf(sv("Error: could not open file '%'"), make<StringView>(file_to_parse));
```
antroseco, line 1403: Should add formatting support to `Cstring` at some point

src/glang/parser/parser.c3
```diff
@@ -1437,8 +1434,5 @@ function main : (argc: int, argv: u8[&][&]) -> int = {
         json_output:write_json_value(&program);
         json_output:close();
     }
-
-    // 4) Cleanup
-    file_contents:deallocate();
```
antroseco, line 1442: Still need to call `unmap`, right?

antroseco (commented):
Super cool!

We'll need a way to auto-generate the platform-specific code, but that's outside the scope of this PR.

src/glang/parser/parser.c3
DaveDuck321: Added! Which is kinda funny when we're leaking literal the entire allocator

src/glang/lib/std/sys/types.c3
DaveDuck321: I've removed these for now. I'm trying to take stat directly from the kernel implementation instead.

src/glang/lib/std/sys/stat.c3
DaveDuck321: Yeah, I was thinking of that... The problem is libc doesn't actually need to match the syscall's memory layout... We'd need to peek into the kernel source to autogenerate.

src/glang/lib/std/sys/stat.c3
DaveDuck321: Ahhh good spot... But we have function overloading, no need to encode the type into the signature :upside_down_face:.

src/glang/lib/std/sys/stat.c3
DaveDuck321: I didn't like it either. I've moved this into `syscalls.c3` to remove one of the casts. This seems like a good idea since the struct layout is already part of the syscall.

src/glang/lib/std/io.c3
DaveDuck321: Hmm, yeah. Maybe this should be a separate commit. It felt like posix was leaking into one of our non-posix headers.

src/glang/lib/std/io.c3
DaveDuck321: ooo, I forgot about this :heart_eyes:

src/glang/lib/std/io.c3
DaveDuck321: It doesn't... Guess we just weren't calling it before hmmm

src/glang/lib/std/io.c3
DaveDuck321: Yeah + it was a bit misleading since it didn't actually buffer the read. :smile:

DaveDuck321:
Yay, arm works.
x86 and x86_64 share kernel headers. Arm and Arm64 don't tho.
libc does whatever it wants and misuses the api so hard that linux `newstatat` has more libc workarounds than code.

DaveDuck321:
I ran a few benchmarks. We now beat the c++ standard library on my computer!! Also interestingly, c++ performs very poorly until we start writing non-ergonomic code. I benchmarked the closer test using a warmup of 5, c++ loses by >10%.

```c++
#include <iostream>
#include <fstream>

int main () {
    std::ifstream infile("test_file.txt");

    size_t sum_of_ones = 0;
    while (infile) {
        char next_char = infile.get();
        if (next_char == '1') {
            sum_of_ones += 1;
        }
    }
    std::cout << "Sum of ones: " << sum_of_ones << std::endl;
    return 0;
}

/*
________________________________________________________
Executed in    3.38 secs    fish           external
   usr time    3.28 secs    0.00 micros    3.28 secs
   sys time    0.09 secs  356.00 micros    0.09 secs
*/
```

```c++
#include <iostream>
#include <fstream>
#include <span>

int main () {
    std::ifstream infile ("test_file.txt");
    infile.seekg(0, std::ios::end);
    const auto size = infile.tellg();
    const auto content = new char [size];
    infile.seekg(0, std::ios::beg);
    infile.read (content, size);
    infile.close();

    size_t sum_of_ones = 0;
    for (char next_char : std::span(content, size)) {
        if (next_char == '1') {
            sum_of_ones+=1;
        }
    }
    delete[] content;

    std::cout << "Sum of ones: " << sum_of_ones << std::endl;
    return 0;
}

/*
________________________________________________________
Executed in  624.55 millis    fish           external
   usr time  252.46 millis    0.00 micros  252.46 millis
   sys time  370.45 millis  303.00 micros  370.15 millis
*/
```

```c++
#include <cstddef>
#include <iostream>
#include <fstream>
#include <span>

int main () {
    std::ifstream infile ("test_file.txt");
    char buffer[1024];

    size_t sum_of_ones = 0;
    while (infile) {
        infile.read(buffer, 1024);
        for (char next_char : std::span(buffer, infile.gcount())) {
            if (next_char == '1') {
                sum_of_ones += 1;
            }
        }
    }

    std::cout << "Sum of ones: " << sum_of_ones << std::endl;
    return 0;
}

/*
329.9 ms ±   2.0 ms    [User: 218.4 ms, System: 110.5 ms]
________________________________________________________
Executed in  356.63 millis    fish           external
   usr time  241.10 millis  203.00 micros  240.90 millis
   sys time  114.59 millis  122.00 micros  114.47 millis
*/
```

```rust
@require_once "std/io.c3"
@require_once "std/format.c3"

function main : () -> int = {
    let maybe_file: Optional<File> = open(make<CString>("test_file.txt\0"), O_RDONLY());
    let file: File& = &maybe_file:data();

    let data : Optional<MappedFile> = file:read();
    if !data:has_value() {
        print(sv("Failed to read file"));
        return 1;
    }
    let mapping : MappedFile& = &data:data();
    file:close();

    mut sum_of_ones : isize = 0;
    for char in mapping.buffer:get_iter() {
        if char == char("1") {
            sum_of_ones += 1;
        }
    }
    mapping:unmap();

    print(sum_of_ones);
    return 0;
}

/*
291.4 ms ±   5.7 ms    [User: 236.7 ms, System: 53.8 ms]
________________________________________________________
Executed in  311.81 millis    fish           external
   usr time  270.65 millis    0.00 micros  270.65 millis
   sys time   40.27 millis  328.00 micros   39.95 millis
*/

```

src/glang/parser/parser.c3
antroseco: We'll get there :)

src/glang/lib/std/sys/stat.c3
antroseco: Fair enough, I'll accept that

src/glang/lib/std/sys/stat.c3
antroseco: That's fine - I don't think getting the kernel headers is any harder than the libc headers

src/glang/lib/std/io.c3
antroseco: Oops!

src/glang/lib/std/io.c3
antroseco: Yeah I wrote `puts` based on POSIX. Still not sure whether we should provide a POSIX-like API or not

src/glang/lib/std/io.c3
antroseco: As you'd expect ;)

src/glang/lib/std/aarch64_linux/sys/kstat.c3
```diff
@@ -1,21 +1,23 @@
-// https://elixir.bootlin.com/linux/latest/source/arch/arm/include/uapi/asm/stat.h
+// https://elixir.bootlin.com/linux/latest/source/include/uapi/asm-generic/stat.h
```
antroseco, line 1:
```
 * Everybody gets this wrong and has to stick with it for all
 * eternity. Hopefully, this version gets used by new architectures
 * so they don't fall into the same traps.
```

:rocket:

antroseco (approved):

antroseco:
> I ran a few benchmarks. We now beat the c++ standard library on my computer!! Also interestingly, c++ performs very poorly until we start writing non-ergonomic code. I benchmarked the closer test using a warmup of 5, c++ loses by >10%.

Damn, nice. Some very interesting results.

Good to see that we're doing something right :)

Also, can I suggest https://github.com/sharkdp/hyperfine for benchmarks like this - it's pretty cool

DaveDuck321:
> Also, can I suggest https://github.com/sharkdp/hyperfine for benchmarks like this - it's pretty cool

Lol, we found the same library. My warmup was the `--warmup` argument. Hyperfine gave the statistics above the unix time :smile:.

antroseco:
> > Also, can I suggest [sharkdp/hyperfine](https://github.com/sharkdp/hyperfine) for benchmarks like this - it's pretty cool
> 
> Lol, we found the same library. My warmup was the `--warmup` argument. Hyperfine gave the statistics above the unix time 😄.

Ah, nice!

PR #143, opened 2024-01-14
Add Map container and RB tree iterator

antroseco:
The goal was to use `Map` in the JSON library, but there's a bunch of stuff in this PR:

- Add `Map`, based on `RB_Tree`
- Fixed the debug information generated for struct members
- Add RB tree `predecessor()` and `successor()` functions
- Add RB tree and Map iterators
- Change some comparisons to take their arguments by references (the only straightforward way to compare two `JSON_Node`s is by looking at their address... but to pass them by reference means that we have to pass everything else by reference too, even `int`s)

### DIFF
--- a/src/glang/codegen/builtin_types.py
+++ b/src/glang/codegen/builtin_types.py
@@ -498,14 +498,11 @@ class StructDefinition(ValueTypeDefinition):
 
     @property
     def size(self) -> int:
-        member_sizes = [member.size for _, member in self.members]
-        member_aligns = [member.alignment for _, member in self.members]
-        return get_abi().compute_struct_size(member_sizes, member_aligns)
+        return get_abi().compute_struct_size(self._member_sizes, self._member_aligns)
 
     @property
     def alignment(self) -> int:
-        member_aligns = [member.alignment for _, member in self.members]
-        return get_abi().compute_struct_alignment(member_aligns)
+        return get_abi().compute_struct_alignment(self._member_aligns)
 
     def get_member_by_name(self, target_name: str) -> tuple[int, Type]:
         for index, (member_name, member_type) in enumerate(self.members):
@@ -513,6 +510,14 @@ class StructDefinition(ValueTypeDefinition):
                 return index, member_type
 
         raise FailedLookupError("struct member", "{" + target_name + " : ... }")
+
+    @property
+    def _member_sizes(self) -> list[int]:
+        return [member.size for _, member in self.members]
+
+    @property
+    def _member_aligns(self) -> list[int]:
+        return [member.alignment for _, member in self.members]
 
     def to_di_type(self, metadata_gen: Iterator[int]) -> list[Metadata]:
         # TODO template parameters (see DWARF5 2.23).
@@ -532,8 +537,13 @@ class StructDefinition(ValueTypeDefinition):
 
         other_metadata: list[Metadata] = []
         di_derived_types: list[DIDerivedType] = []
-        curr_offset = 0
-        for m_name, m_type in self.members:
+        for (m_name, m_type), m_offset in zip(
+            self.members,
+            get_abi().compute_struct_member_offsets(
+                self._member_sizes, self._member_aligns
+            ),
+            strict=True,
+        ):
             base_type_metadata = m_type.to_di_type(metadata_gen)
             other_metadata.extend(base_type_metadata)
             assert isinstance(base_type_metadata[-1], DIType)
@@ -545,11 +555,9 @@ class StructDefinition(ValueTypeDefinition):
                     8 * m_type.size,
                     Tag.member,
                     base_type_metadata[-1],
-                    8 * curr_offset,
+                    8 * m_offset,
                 )
             )
-
-            curr_offset += m_type.size
 
         metadata_list = MetadataList(next(metadata_gen), di_derived_types)
         self._di_type.elements = metadata_list
--- a/src/glang/lib/std/arithmetic.c3
+++ b/src/glang/lib/std/arithmetic.c3
@@ -9,6 +9,7 @@ typedef [U, V] AreCompatibleArithmeticTypes :
     >
 
 typedef [U, V] CommonArithmeticType: TypeIf<U, AreCompatibleArithmeticTypes<U, V>>
+typedef [U, V] CommonArithmeticType<U&, V&>: TypeIf<U, AreCompatibleArithmeticTypes<U, V>>
 
 typedef CommonArithmeticType<i8, i16>   : i16
 typedef CommonArithmeticType<i8, i32>   : i32
@@ -29,6 +30,7 @@ typedef CommonArithmeticType<f32, f128> : f128
 typedef CommonArithmeticType<f64, f128> : f128
 
 typedef [U, V] ArithmeticCompareResult : TypeIf<bool, AreCompatibleArithmeticTypes<U, V>>
+typedef [U, V] ArithmeticCompareResult<U&, V&> : TypeIf<bool, AreCompatibleArithmeticTypes<U, V>>
 typedef [T] RegisterHasTrivialAssignments<Override<T>> : TypeIf<TrueType, IsArithmetic<T>>
 
 @operator [T] + : (value: T) -> TypeIf<T, IsArithmetic<T>> = { return value; }
--- a/src/glang/lib/std/json.c3
+++ b/src/glang/lib/std/json.c3
@@ -1,10 +1,11 @@
 @require_once "std/arithmetic.c3"
+@require_once "std/iterators.c3"
 @require_once "std/logical.c3"
-@require_once "std/string.c3"
+@require_once "std/map.c3"
 @require_once "std/memory.c3"
+@require_once "std/string.c3"
 @require_once "std/vector.c3"
 @require_once "std/wrappers.c3"
-@require_once "std/iterators.c3"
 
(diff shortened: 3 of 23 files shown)


src/glang/lib/std/map.c3
```diff
@@ -0,0 +1,102 @@
...
+}
+
+function [K, V, C] contains : (self : Map<K, V>&, key : C) -> bool = {
+    let result : Optional<V> = self:at(key);
+    return result:has_value();
+}
+
+function [K, V] get_citer : (map : Map<K, V> mut&) -> MapIter<K, V> = {
+    return map.tree:get_citer();
+}
```
DaveDuck321, line 102: Not really a fan of this. Seems very isoc++. Maybe something caller side like (&map):get_iter()?

src/glang/lib/std/json.c3
```diff
@@ -449,22 +456,15 @@ function [IO] write : (io : IO mut&, node : JSON_Node) -> TypeIf<StringView, IsS
...
         for element in &mut iter {
-            io:write(element);
-
-            if flag {
-                io:write(sv(": "));
-            } else {
-                if iter:has_next() {
-                    io:write(sv(", "));
-                }
+            io:writef(sv("%: %"), element.key, element.value);
```
DaveDuck321, line 464: ~We should be quoting + escaping element keys right? That way we emit valid json as well as parsing it.~

Ohh, both the key and a value are a json_node, nvm then. This could be optimized to fix all keys as a string but thats a nit

src/glang/lib/std/map.c3
```diff
@@ -0,0 +1,102 @@
...
+
+typedef [K, V] MapNode : {
+    key   : K,
+    value : V
+}
+
+typedef [K, V] MapIter : RB_Iter<MapNode<K, V>>
+
+// FIXME <=> is untested and broken :(
+typedef [K, V] HasSpaceship<MapNode<K, V>> : HasSpaceship<K>
```
DaveDuck321, line 15: :cry: Have you started working on it?

src/glang/lib/std/map.c3
```diff
@@ -0,0 +1,102 @@
...
+
+@operator [K1, K2, V] == : (left : MapNode<K1, V>&, right : MapNode<K2, V>&) -> bool = {
+    return &left.key == &right.key;
+}
+
+@operator [K1, K2, V] != : (left : MapNode<K1, V>&, right : MapNode<K2, V>&) -> bool = {
+    return &left.key != &right.key;
+}
```
DaveDuck321, line 23: Fortunately, the strange syntax stops being awkward when (if) we add automatic const ref promotion.

tests/stdlib/map.c3
```diff
@@ -0,0 +1,73 @@
...
+    print(map:at(2));
+    print(map:at(1));
+    print(map:at(0));
+
+    for i in map:get_citer() {
+        printf(sv("% = %"), i.key, i.value);
+    }
+
+    runtime_assert(map:erase(0));
+    runtime_assert(!map:erase(4));
```
DaveDuck321, line 36: Wanna change `!` to `not`. It'd match our `and`/ `or` syntax?

src/glang/lib/std/vector.c3
```diff
@@ -66,43 +66,19 @@ function [T] get : (self : Vector<T>&, index : isize) -> T mut& = {
...
 }
 
-function [T] capacity : (self : Vector<T> mut&) -> isize = {
-    return self.memory.length;
-}
-
-function [T] data : (self : Vector<T> mut&) -> Span<T> = {
-    return self.memory:slice_to(self.length);
-}
-
```
DaveDuck321, line 94: Good to see this go

src/glang/lib/std/json.c3
```diff
@@ -132,8 +141,6 @@ function parse_object : (cursor : Cursor mut&) -> Optional<JSON_Node> = {
             return make<Optional<JSON_Node>>();
```
DaveDuck321, line 141: I can't comment the previous line but you're still deallocating the vector, this should be updated to the map

DaveDuck321 (changes requested):
`map` is something I've wanted for a long time. This is really nice.

While not the fault of this PR, I spent a few minutes here trying to understand what's allocating and when. My understanding is that 1) the json parser creates a new allocator which is shared among all nodes (used for their vectors). 2) Each node has a map: in the map constructor, a rb_tree is constructed, which initializes a new allocator -- each node gets 1 page!?

Please can we just never call `initialize_allocator()` in the standard library and just pass around the user's allocator? At least until we work out the ergonomics of a global allocator (which I'm not convinced is even a good idea -- why most embedded c programmer say c++ bad)

src/glang/lib/std/arithmetic.c3
```diff
@@ -29,6 +30,7 @@ typedef CommonArithmeticType<f32, f128> : f128
 typedef CommonArithmeticType<f64, f128> : f128
 
 typedef [U, V] ArithmeticCompareResult : TypeIf<bool, AreCompatibleArithmeticTypes<U, V>>
+typedef [U, V] ArithmeticCompareResult<U&, V&> : TypeIf<bool, AreCompatibleArithmeticTypes<U, V>>
```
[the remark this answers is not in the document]
antroseco: I don't really like it either, but it'll do until we decide how operators and borrowing is going to work. Maybe take some inspiration from rust.

src/glang/codegen/builtin_types.py
```diff
@@ -532,8 +537,13 @@ def to_di_type(self, metadata_gen: Iterator[int]) -> list[Metadata]:
...
         di_derived_types: list[DIDerivedType] = []
-        curr_offset = 0
-        for m_name, m_type in self.members:
+        for (m_name, m_type), m_offset in zip(
+            self.members,
+            get_abi().compute_struct_member_offsets(
+                self._member_sizes, self._member_aligns
+            ),
+            strict=True,
+        ):
```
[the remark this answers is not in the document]
antroseco: Yeah, I think I just completely forgot about padding when I wrote it the first time...

src/glang/lib/std/json.c3
```diff
@@ -21,18 +22,26 @@ typedef JSON_Node : {
...
         node_type,
         make<Optional<StringView>>(),
         make<Vector<JSON_Node>>(&mut allocator),
+        make<Map<JSON_Node, JSON_Node>>(),
     };
 }
 
+@operator < : (left : JSON_Node&, right : JSON_Node&) -> bool = {
+    return ref_to_addr(&left) < ref_to_addr(&right);
+}
```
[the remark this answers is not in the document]
antroseco: It's just tedious... I think I'll refactor it to use a `StringView` instead of a `JSON_Node`, then we only have to write `strcmp`.

src/glang/lib/std/map.c3
antroseco: Hmm, not sure what you mean here

src/glang/lib/std/map.c3
antroseco: Not yet, feel free to pick it up

tests/stdlib/map.c3
antroseco: I wouldn't be opposed to it. Should we make it non-overloadable then? To match the convention of `and`/`or`?

Also, it'd be pretty strange to change the meaning of the keyword `not`

DaveDuck321: Yeah, idm that. It's not like an overload of `!` makes sense anyway

src/glang/lib/std/json.c3
```diff
@@ -21,18 +22,26 @@ typedef JSON_Node : {
                     // 5: array, 6: object
     value    : Optional<StringView>,
     children : Vector<JSON_Node>,
+    map      : Map<JSON_Node, JSON_Node>,
```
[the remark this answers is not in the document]
antroseco: One day...

src/glang/lib/std/json.c3
antroseco: Done, and added a FIXME for the escaping bit

src/glang/lib/std/json.c3
antroseco: Oops, fixed

antroseco:
> `map` is something I've wanted for a long time. This is really nice.
> 
> While not the fault of this PR, I spent a few minutes here trying to understand what's allocating and when. My understanding is that 1) the json parser creates a new allocator which is shared among all nodes (used for their vectors). 2) Each node has a map: in the map constructor, a rb_tree is constructed, which initializes a new allocator -- each node gets 1 page!?
> 
> Please can we just never call `initialize_allocator()` in the standard library and just pass around the user's allocator? At least until we work out the ergonomics of a global allocator (which I'm not convinced is even a good idea -- why most embedded c programmer say c++ bad)

Ugh, yeah that's fair. I've removed all calls - now we need to pass an allocator to `parse_json()` instead.

src/glang/lib/std/map.c3
DaveDuck321: Oh, rip. Maybe my fault then. Possibly even copilot

DaveDuck321 (approved):
Cool, nice to see this go.

src/glang/lib/std/map.c3
antroseco: Could always change it

PR #144, opened 2024-01-17
Change return type of `write()` to void

antroseco:
Also fixed type-checking when returning `void` from a function (we weren't doing any...)

### DIFF
--- a/tests/regressions/return_void.c3
+++ b/tests/regressions/return_void.c3
@@ -1,0 +1,7 @@
+function main : () -> int = {
+    return;
+}
+
+/// @COMPILE; EXPECT ERR
+/// File '*.c3', line 2, in 'function main: () -> int'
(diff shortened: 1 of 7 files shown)


tests/regressions/return_void.c3
```diff
@@ -0,0 +1,7 @@
+function main : () -> int = {
+    return;
+}
+
+/// @COMPILE; EXPECT ERR
+/// File '*.c3', line 2, in 'function main: () -> int'
+/// Error in return statement: type 'int' does not match expected type 'void'
```
DaveDuck321, line 7: I've defo encountered this before. Could have sworn I opened a bug... But I guess not. Thanks for fixing it :heart:

DaveDuck321 (approved):
Very nice. Perfect change. This API is much more robust now.

PR #150, opened 2024-03-11
Clean up with ruff

antroseco:
This is very gratuitous, but I was playing around with https://github.com/astral-sh/ruff.

### DIFF
--- a/src/glang/codegen/__init__.py
+++ b/src/glang/codegen/__init__.py
@@ -1,4 +1,4 @@
-from .builtin_types import (
+from glang.codegen.builtin_types import (
     BoolType,
     FunctionSignature,
     IEEEFloat,
@@ -7,8 +7,8 @@ from .builtin_types import (
     UIntType,
     VoidType,
 )
-from .debug import DIFile
-from .expressions import (
+from glang.codegen.debug import DIFile
+from glang.codegen.expressions import (
     ArrayIndexAccess,
     BorrowExpression,
     ConstantExpression,
@@ -19,7 +19,7 @@ from .expressions import (
     UnnamedInitializerList,
     VariableReference,
 )
-from .generatable import (
+from glang.codegen.generatable import (
     Assignment,
     IfElseStatement,
     ReturnStatement,
@@ -28,7 +28,7 @@ from .generatable import (
     VariableInitialize,
     WhileStatement,
 )
-from .interfaces import (
+from glang.codegen.interfaces import (
     Generatable,
     GenericArgument,
     GenericMapping,
@@ -37,8 +37,8 @@ from .interfaces import (
     TypedExpression,
     Variable,
 )
-from .translation_unit import Function, Program, Scope
-from .type_resolution import (
+from glang.codegen.translation_unit import Function, Program, Scope
+from glang.codegen.type_resolution import (
     CompileTimeConstant,
     FunctionDeclaration,
     GenericValueReference,
@@ -55,3 +55,55 @@ from .type_resolution import (
     UnresolvedType,
     UnresolvedTypeWrapper,
 )
+
+__all__ = [
+    "ArrayIndexAccess",
+    "Assignment",
+    "BoolType",
+    "BorrowExpression",
+    "CompileTimeConstant",
+    "ConstantExpression",
+    "DIFile",
+    "Function",
+    "FunctionDeclaration",
+    "FunctionSignature",
+    "Generatable",
+    "GenericArgument",
+    "GenericMapping",
+    "GenericValueReference",
+    "IEEEFloat",
+    "IfElseStatement",
+    "InitializerList",
+    "IntType",
+    "LogicalOperator",
+    "NamedInitializerList",
+    "NumericLiteralConstant",
+    "Program",
+    "ReturnStatement",
+    "Scope",
+    "SpecializationItem",
+    "StackVariable",
+    "StaticVariable",
+    "StructDefinition",
+    "StructMemberAccess",
+    "Type",
+    "TypedExpression",
+    "Typedef",
+    "UIntType",
+    "UnnamedInitializerList",
+    "UnresolvedFunctionSignature",
+    "UnresolvedGenericType",
+    "UnresolvedHeapArrayType",
+    "UnresolvedNamedType",
+    "UnresolvedReferenceType",
+    "UnresolvedSpecializationItem",
+    "UnresolvedStackArrayType",
+    "UnresolvedStructType",
+    "UnresolvedType",
+    "UnresolvedTypeWrapper",
+    "Variable",
+    "VariableInitialize",
+    "VariableReference",
+    "VoidType",
+    "WhileStatement",
+]
--- a/src/glang/codegen/builtin_types.py
+++ b/src/glang/codegen/builtin_types.py
@@ -6,8 +6,7 @@ from functools import cached_property, reduce
 from operator import mul
 from typing import Optional
 
-from ..target import get_abi, get_int_type_info, get_ptr_type_info
-from .debug import (
+from glang.codegen.debug import (
     DIBasicType,
     DICompositeType,
     DIDerivedType,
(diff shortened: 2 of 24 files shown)


src/glang/codegen/__init__.py
```diff
@@ -55,3 +55,55 @@
     UnresolvedType,
     UnresolvedTypeWrapper,
 )
+
+__all__ = [
```
DaveDuck321, line 59: This seems a bit redundant and error-prone to me... but also if this is automatically generated by ruff I don't mind.

src/glang/codegen/builtin_types.py
```diff
@@ -271,18 +272,18 @@ def are_equivalent(self, other: TypeDefinition) -> bool:
 
     @property
     def size(self) -> int:
-        assert False
+        raise AssertionError
```
DaveDuck321, line 279: Huh, why isn't this instantiating the AssertionError? Is this not a class?

src/glang/codegen/expressions.py
```diff
@@ -386,11 +386,10 @@ def __init__(
...
                 )
-        else:
-            if len(indices) != 1 + len(array_definition.known_dimensions):
-                raise ArrayIndexCount(
-                    self._type_of_array.format_for_output_to_user(), len(indices), 1
-                )
+        elif len(indices) != 1 + len(array_definition.known_dimensions):
+            raise ArrayIndexCount(
+                self._type_of_array.format_for_output_to_user(), len(indices), 1
+            )
```
DaveDuck321, line 389: Impressive pattern match. Good job ruff

src/glang/codegen/type_conversions.py
```diff
@@ -197,7 +197,7 @@ def last_type() -> Type:
                 context,
                 src.format_for_output_to_user(),
                 dest_type.format_for_output_to_user(True),
-                maybe_missing_borrow,
+                maybe_missing_borrow=maybe_missing_borrow,
```
DaveDuck321, line 198: I just don't like this. The type name adds nothing caller-side.

Rust allows both specifically in this case :upside_down_face: .

But I like *almost* everything else ruff is doing so I won't complain if this is automated.

src/glang/codegen/type_resolution.py
```diff
@@ -431,7 +434,9 @@ def pattern_match(
             return None
 
         cost = 0
-        for our_member, target_member in zip(self.members, target.definition.members):
+        for our_member, target_member in zip(
+            self.members, target.definition.members, strict=True
+        ):
```
DaveDuck321, line 433: At this point we might as well add our own `zips` helper

src/glang/codegen/builtin_types.py
```diff
@@ -81,15 +86,15 @@ def __init__(
         name: str,
         specialization: list[SpecializationItem],
         definition: TypeDefinition,
-        alias: Optional[Type],
+        alias: Type | None,
```
DaveDuck321, line 89: Hmm, I'm not convinced... Maybe I'll warm to it

src/glang/codegen/generatable.py
```diff
@@ -1,18 +1,22 @@
-from typing import Iterable, Optional
+from __future__ import annotations
```
DaveDuck321, line 1: Can't we just require a recent python version for this?
Or does this add something else that isn't in 3.10+?

src/glang/codegen/type_conversions.py
```diff
@@ -227,8 +225,8 @@ def last_type() -> Type:
     if (
         last_type().storage_kind.is_reference()
         and last_type().storage_kind == dest_type.storage_kind
-        and isinstance(last_array_def, (HeapArrayDefinition, StackArrayDefinition))
-        and isinstance(dest_array_def, (HeapArrayDefinition, StackArrayDefinition))
+        and isinstance(last_array_def, HeapArrayDefinition | StackArrayDefinition)
+        and isinstance(dest_array_def, HeapArrayDefinition | StackArrayDefinition)
```
DaveDuck321, line 229: Oh wow, cool syntax

src/glang/target.py
```diff
@@ -83,7 +82,7 @@ def compute_padding(curr_size: int, align_to: int) -> int:
             # Append member to the struct
             struct_size += member_size
 
-        # The size of any object is always a multiple of the object‘s alignment.
```
DaveDuck321, line 86: How did we even sneak the non-asic ‘ in here?

src/glang/codegen/builtin_types.py
DaveDuck321: > Otherwise, [raise](https://docs.python.org/3/reference/simple_stmts.html#raise) evaluates the first expression as the exception object. It must be either a subclass or an instance of [BaseException](https://docs.python.org/3/library/exceptions.html#BaseException). If it is a class, the exception instance will be obtained when needed by instantiating the class with no arguments.

Oh god. This is jank

tests/test_config_parser.py
```diff
@@ -110,10 +111,11 @@ def grep_ir_cmd(self, token: Token) -> None:
 
 def parse_file(path: Path) -> TestConfig:
     with path.open(encoding="utf-8") as file:
-        lines = map(
-            lambda line: line.removeprefix("///").strip(),
-            filter(lambda line: line.startswith("///"), file.readlines()),
-        )
+        lines = [
+            line.removeprefix("///").strip()
+            for line in file.readlines()
```
DaveDuck321, line 116: Not 100% sure but its in the blame now :upside_down_face: 
```suggestion
            for line in file
```

src/glang/graphene_parser.py
```diff
@@ -22,19 +23,21 @@
...
     VoidVariableDeclaration,
 )
-from .parser import lexer_parser as lp
+from glang.parser import lexer_parser as lp
 
 UnresolvedGenericMapping = dict[str, cg.UnresolvedSpecializationItem]
 
 
 class ResolvedPath(str):
+    __slots__ = ()
```
DaveDuck321, line 32: Can you explain why we'd want this?

DaveDuck321 (commented):
Might as well add ruff to the CI now?

I'm not a fan of every change (looking at you janky exception auto-initialization) but if it were mandatory I probably wouldn't care

src/glang/codegen/__init__.py
antroseco: Yeah it's all auto-generated, but most importantly it stops pylance from thinking that all the imports are unused

src/glang/codegen/builtin_types.py
antroseco: Yeah, I can't believe this is a feature...

src/glang/codegen/type_conversions.py
antroseco: > Rust allows both specifically in this case 🙃 .

Oof, rust has all the features.

> But I like almost everything else ruff is doing so I won't complain if this is automated.

This one isn't actually (not everything is automated), but we could disable that rule

src/glang/codegen/type_resolution.py
antroseco: Idk, sometimes we do actually want `strict=False` (but really the default should be true)

src/glang/codegen/builtin_types.py
antroseco: lol, we can change that back if you want

src/glang/codegen/generatable.py
antroseco: There's no version of Python with that feature yet, see https://docs.python.org/3/library/__future__.html#module-contents

It just means that we don't have to wrap type annotations in strings any more

src/glang/codegen/type_conversions.py
antroseco: Turns out, Python has all the features too

src/glang/target.py
antroseco: I think I just copy-pasted that from the spec, wooops

src/glang/graphene_parser.py
antroseco: https://docs.astral.sh/ruff/rules/no-slots-in-str-subclass/

I'll add that as a comment

antroseco:
> Might as well add ruff to the CI now?
> 
> I'm not a fan of every change (looking at you janky exception auto-initialization) but if it were mandatory I probably wouldn't care

Thing is, there's still a few hundred rule violations in the code base lol

Most of the changes in this diff were done automatically, but some stuff I did manually cause I thought it was interesting.

We could add this to the CI but only keep the rules with automatic fixes, and maybe a very small number of rules that require manual changes.

src/glang/codegen/builtin_types.py
DaveDuck321: It fine. You've made the change now. Is it caught by the linter or was this a manual search?

src/glang/codegen/generatable.py
DaveDuck321: Fair. Seems reasonable

src/glang/codegen/builtin_types.py
antroseco: It's caught by the linter

tests/test_config_parser.py
antroseco: And it works!

antroseco:
@DaveDuck321 I disable all the rules without automatic fixes from the `pyproject.toml` and added ruff to the CI

PR #154, opened 2024-03-24
Implement the `not` operator

antroseco:
Adds support for a built-in logical not. I can't remove the existing `operator !` just yet, because bootstrap, so we have to support both for now.


src/glang/parser/parser.c3
```diff
@@ -974,7 +975,16 @@ function parse_non_operator_expression_prefix : (cursor : Cursor mut&) -> Expres
...
+    if cursor:try_skip_next_and_surrounding_white_space("not") {
+        let rhs : Expression = cursor:parse_maybe_unary_or_borrow_expression();
+
+        let operator_call : UnaryLogicalOperator mut& = &mut cursor.allocator:allocate<UnaryLogicalOperator>();
+        operator_call = {{start_offset, start_offset + length("not")}, rhs, meta(start_offset, rhs.meta.end_offset)};
+
+        return {13, ref_to_addr(&operator_call), operator_call.meta};
+    }
+
+    if is_one_of(cursor:get_char(), char("!"),char("-"), char("+"), char("~")) {
```
DaveDuck321, line 987: Hmm, this has the same behaviour as before... but also this is slightly misleading now right?

I'd expect: `not a == b` to evaluate as `not (a == b)` but with this change it evaluates as `(not a) == b`... This breaks if `a` or `b` isn't a bool... but also I can't think of any other situation it'd be a problem, so probably ok?

antroseco: Right, it behaves exactly like `!` used to, just spelt out.

I was about to say this isn't really misleading as that's what we're used to anyway, and then I thought I'd check what Python does... and well that's quite hard...

```
not True == False
```

as either evaluation produces the same result, duh

I guess the only case where this matters is when the two things are not bools, like you said, so we might as well reduce the precedence of `not` and get that to work

PR #160, opened 2024-06-19
Driver improvements: linker scripts + asm output

DaveDuck321:


src/glang/driver.py
```diff
@@ -284,6 +313,13 @@ def main() -> None:
     else:
         opt_sources = [llvm.optimize(args) for llvm in llvm_sources]
 
+    # 2.5) Emit asm
+    if args.emit_asm or args.emit_everything:
+        for source in opt_sources:
+            asm = source.assemble(args)
+            target_output = asm.get_top_level_source_file().with_suffix(".S")
```
antroseco, line 320: nit, but I prefer `.s`
- consistent with gcc and clang
- `.S` generally means the file needs to go through the C preprocessor before assembly

antroseco (approved):

PR #161, opened 2024-08-26
Fix bootstrap broken by lark 1.2.2

DaveDuck321:
Lark 1.2.2 was release on August 13. This was marked as backwards compatible with lark 1.1 (hence pulled in by Hatch). Turns out it is not in fact backwards compatible and was breaking our bootstrap on a fresh checkout.

This PR reverts lark to the latest backwards compatible version.


DaveDuck321:
Looks like they won't fix it on their end: https://github.com/lark-parser/lark/issues/1456

antroseco (commented):
Ah, that sucks

PR #162, opened 2024-08-27
Test on Python 3.12 and 3.13

antroseco:
`actions/setup-python` now supports architectures other than x86/x64, so we can start using it again

### DIFF
--- a/.github/workflows/graphene_tests.yml
+++ b/.github/workflows/graphene_tests.yml
@@ -10,7 +10,7 @@ jobs:
     strategy:
       matrix:
         llvm-version: [17, 18]
-        python-version: [3.12, 3.13]
+        python-version: [3.12, 3.13.0-rc.1]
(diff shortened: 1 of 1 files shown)


DaveDuck321 (approved):

.github/workflows/graphene_tests.yml
```diff
@@ -10,6 +10,7 @@ jobs:
     strategy:
       matrix:
         llvm-version: [17, 18]
+        python-version: [3.12, 3.13.0-rc.1]
```
DaveDuck321, line 13: Actually shall we just trust that python is forwards compatible? This seems like it'll double CI compute for quite a small benefit.

The proper action support is cool tho.

antroseco: Sure, although it's not always forwards compatible. We don't do anything weird so it's likely fine.

PR #163, opened 2024-08-28
Testing: add `@FAILING` tag

DaveDuck321:
I've been working on function pointers, callables, and an over-loadable [] operator... Unfortunately this is becoming unwieldy and I'll have to regress the multidimensional array test to avoid touching the pattern matching logic in the same diff.

This PR adds the testing infrastructure to make this possible without completely deleting the test.

### DIFF
--- a/tests/run_tests.py
+++ b/tests/run_tests.py
@@ -1,5 +1,6 @@
 import subprocess
 from concurrent.futures import ThreadPoolExecutor
+from enum import Enum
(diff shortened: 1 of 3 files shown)


antroseco (commented):
What if we called it `@FAILING` instead, and then _checked_ that it failed? You'd achieve the same result, but then you can no longer "forget" to re-enable the tests.

antroseco (approved):

PR #164, opened 2024-08-28
Support array index operator overloading

DaveDuck321:
I started this a few days ago with the goal of adding function pointer and callable support. It kinda snowballed into something completely different.

This PR makes the following changes:
1) We can overload the `[]` operator
2) Builtin arrays no longer know how to index, they require an operator definition (like our ints)
3) Double reference errors now trigger a substitution failure

Reasoning:
1) This is the same mechanism we will use to overload the `()` operator so I thought it'd be clean to test this first.
2) This let me reuse the variadic support and template magic we have for functions.
3) This is required for change 2 to correctly match even with the `Ref<>` helper

Since I'm introducing a bootstrap stage for this anyway, I made the following fixes:
1) Overloads for `<=>` will parse
2) Function types will parse
3) Overloads for `()` will parse

Regression:
- This regresses the multidimensional array test since we can't pattern match like this atm: `T[mut&, @N...]` therefore we can't write an array index overload that will for any size of array.


bootstrap.sh
```diff
@@ -47,7 +52,17 @@ env -C "$stage_4_dir" python3 -m src.glang.driver ./src/glang/parser/parser.c3 -
 
 # Stage 5; we introduce floating point literal support, compile the last parser
 # that supports this before literals become required by the standard library.
-env -C "$stage_5_dir" python3 -m src.glang.driver ./src/glang/parser/parser.c3 -o "$dest_dir/dist/parser"
+env -C "$stage_5_dir" python3 -m src.glang.driver ./src/glang/parser/parser.c3 -o "$stage_5_dir/dist/parser"
```
antroseco, line 55:
```suggestion
env -C "$stage_5_dir" python3 -m src.glang.driver ./src/glang/parser/parser.c3 -o "$stage_6_dir/dist/parser"
```
Shouldn't you be building the _next_ stage's parser? And then you can change the Stage 6 command to use `env -C "$stage_6_dir"`

src/glang/codegen/builtin_callables.py
```diff
@@ -606,6 +609,111 @@ def assert_can_write_to(self) -> None:
...
+        self._indices: list[TypedExpression] = []
+        for index in indices:
+            index_expr, conversions = do_implicit_conversion(
+                index, SizeType(), "array index access"
+            )
+            self._indices.append(index_expr)
+            self._conversion_exprs.extend(conversions)
+
+        result_type = array_ptr.result_type_as_if_borrowed
+        StaticTypedExpression.__init__(
```
antroseco, line 651: This feels needlessly error prone; can we not use `super()` instead? (I guess we'll need to inherit from `ABC` instead of making the `BuiltinCallable` constructor abstract, hopefully that works)

src/glang/codegen/builtin_callables.py
```diff
@@ -606,6 +609,111 @@ def assert_can_write_to(self) -> None:
...
+        )
+
+    def generate_ir(self, ctx: IRContext) -> IROutput:
+        # https://llvm.org/docs/LangRef.html#getelementptr-instruction
+        array_def = self._type_of_array.definition
+        assert isinstance(array_def, StackArrayDefinition | HeapArrayDefinition)
+
+        ir = self.expand_ir(self._conversion_exprs, ctx)
+
+        # To access a stack array, we need to dereference the pointer returend
```
antroseco, line 665:
```suggestion
        # To access a stack array, we need to dereference the pointer returned
```

src/glang/lib/std/type_traits.c3
```diff
@@ -23,12 +23,11 @@ typedef [V] AreEquivalent<V, V> : TrueType
 typedef [U, V] SelectFirst : U
 typedef [Type, Condition] TypeIf : SelectFirst<Type, EnableIf<Condition>>
 
-typedef [T] Ref : T&
-typedef [T] Ref<T&> : T&
-typedef [T] Ref<T mut&> : T&
+typedef [T] Ref : TypeIf<T&, Not<IsRef<T>>>
+typedef [T] Ref : TypeIf<T, IsRef<T>>
```
antroseco, line 27: Does this result in a `Y&` if `T = Y mut&`?

src/glang/parser/parser.c3
```diff
@@ -606,6 +636,15 @@ function parse_struct_type : (cursor : Cursor mut&) -> StructType = {
     return {members, meta(start_offset, cursor.position)};
 }
 
+function parse_function_type : (cursor : Cursor mut&) -> FunctionType = {
+    let start_offset : isize = cursor.position;
+
+    let args : Vector<Type> = cursor:parse_type_list_between(sv("("), sv(")"));
+    runtime_assert(cursor:try_skip_next_and_surrounding_white_space("->"));
```
antroseco, line 643: Shouldn't this be an actual syntax error?

tests/stdlib/vector.c3
```diff
@@ -2,6 +2,10 @@
 @require_once "std/vector.c3"
 @require_once "std/format.c3"
 
+@operator[T] [] : (self : Vector<T>&, index : isize) -> T mut& = {
```
antroseco, line 5: This is so nice that it should be in `std/vector.c3`

antroseco (commented):
Super

src/glang/parser/parser.c3
DaveDuck321: Whoops, forgot we actually supported this.

tests/stdlib/vector.c3
DaveDuck321: Yup, this'll break bootstrap tho. We need to wait a commit

src/glang/lib/std/type_traits.c3
DaveDuck321: It would have. I've added a test and updated the traits namse to `ConstRef`, `MutRef`, and `IsAnyRef`

bootstrap.sh
DaveDuck321: The parser binary produced by stage 5 in incompatible with the python changes introduced in stage 6 (since the JSON output format has changed).

The highlighted stage fully bootstraps the stage 5 compiler (before it was in an intermediate state since it was running the stage 4 parser with the stage 5 python). This gives it just enough features to compile the new stage 6 parser with the magic file override.

Since this was the final bootstrap stage there was no need to run a fully bootstrapped stage 6 compiler

DaveDuck321: The next stage looks awful since it was building the stage 6 parser+standard library with the stage 5 python. Note: the stage 6 python can't understand the old JSON format so it can't use the parser compiled as part of stage 5.

antroseco: Yeah, that makes sense. I don't like it (or how convoluted the bootstrap process is) but it makes sense.

tests/stdlib/vector.c3
antroseco: Ah, of course

antroseco (approved):

PR #165, opened 2024-08-31
stdlib: add unicode coding support

DaveDuck321:
This PR adds `UTF8DecodeIter` and the `UTF16DecodeIter` structs for decoding utf-8/16 streams with automatic endianness detection. The goal here is to implement the decoding [standard](https://www.unicode.org/versions/Unicode15.0.0/UnicodeStandard-15.0.pdf) with 100% correctness -- while providing robust (non-fatal) error handling.

Is is possible to produce *some* illegal code point streams in utf16 (corresponding to out-of-order surrogate pairs) and *many* illegal streams in utf8 (corresponding to non-minimal encodings, utf16 surrogate regions, and truncated legal sequences).

Interestingly, correctly detecting errors and converting the offending slices to `�` is actually a required security feature of utf8... Imagine the following SQL injection: `{ <start code-point of a 3-byte utf8 string>, ascii<">, ascii<;> }`.

If an incorrect utf8-based sanitizer parsed this as: `{�}` then it wouldn't know to escape the quote (which in this case was considered part of the invalid utf8 string). The result of this is that when the original string (which passed sanitization) is fed into a simple ASCII-based parser, the ascii parser would see an unknown character, the quote, and then the semicolon. 

To make it safe to mismatch ASCII and UTF-8 parsers, the UTF-8 parser should parse it as: `{�, ascii<">, ascii<;>}` -- ignoring the length (which is normally encoded by the first byte).

Even though its not strictly necessary (utf-8 is well designed :relaxed:), I'm probably going to convert the JSON/ graphene parsers to this new iterator in a follow-up PR.


src/glang/lib/std/string.c3
```diff
@@ -32,7 +32,7 @@ function length : (sv : StringView&) -> isize = {
 }
 
 // TODO: return `UTF32Codepoint` here
```
DaveDuck321, line 34: Intentionally ignored since I don't want to break the parser and conflict with myself rn.

src/glang/lib/std/unicode.c3
```diff
@@ -0,0 +1,412 @@
...
+    {
+        buffer:get(0) = Narrow<u8>(0xe0 | ((scalar.value >> 12) & 0x0f));
+        buffer:get(1) = Narrow<u8>(0x80 | ((scalar.value >> 6) & 0x3f));
+        buffer:get(2) = Narrow<u8>(0x80 | (scalar.value & 0x3f));
+        return {&buffer.data, 3};
+    }
+
+    if is_in_range(scalar.value, 0x10000, 0x3ffff)
+        or is_in_range(scalar.value, 0x40000, 0xfffff)
+        or is_in_range(scalar.value, 0x100000, 0x10ffff)
```
antroseco, line 346: Is `0x1f0000` invalid? Table 3-6 gives this pattern `000uuuuu zzzzyyyy yyxxxxxx`.

src/glang/lib/std/unicode.c3
```diff
@@ -0,0 +1,412 @@
...
+    let first_byte : u8 = self.underlying:__builtin_get_next();
+
+    if !self.underlying:has_next() { return make<Optional<u16>>(); }
+    let second_byte : u8 = self.underlying:__builtin_get_next();
+
+    if self.is_little_endian {
+        return make<Optional<u16>>(first_byte | (as_u16(second_byte) << Narrow<i16>(8)));
+    } else {
+        return make<Optional<u16>>(second_byte | (as_u16(first_byte) << Narrow<i16>(8)));
+    }
```
antroseco, line 377: Should just add a byte swap function in the standard library

src/glang/lib/std/unicode.c3
```diff
@@ -0,0 +1,412 @@
...
+        return { codepoint1 };
+    }
+
+    // None of above, it must be the first part of a surrogate pair
+    let maybe_codepoint2 : Optional<u16> = self:get_next_codepoint();
+    if !maybe_codepoint2:has_value() {
+        return invalid_unicode_scalar();
+    }
+
+    let codepoint2 : u16 = maybe_codepoint2:data();
```
antroseco, line 404: Should we not check that this matches `110111xxxxxxxxxx`?

src/glang/lib/std/util.c3
```diff
@@ -63,6 +63,18 @@ function as_logical : (value : i32)  -> u32  = { return __builtin_bitcast<u32>(v
 function as_logical : (value : i64)  -> u64  = { return __builtin_bitcast<u64>(value); }
 function as_logical : (value : i128) -> u128 = { return __builtin_bitcast<u128>(value); }
 
+function as_i8   : (value : i8)   -> i8  = { return value; }
+function as_i16  : (value : i16)  -> i16  = { return value; }
+function as_i32  : (value : i32)  -> i32  = { return value; }
+function as_i64  : (value : i64)  -> i64  = { return value; }
+function as_i128 : (value : i128) -> i128 = { return value; }
```
antroseco, line 70: Clever

tests/stdlib/unicode/unicode.c3
```diff
@@ -0,0 +1,76 @@
+@require_once "std/format.c3"
```
antroseco, line 1: Would be nice to test some error cases/invalid codepoints as well. Make sure we are substituting the maximal subparts correctly.

antroseco (commented):
Very clean!

src/glang/lib/std/unicode.c3
```diff
@@ -0,0 +1,412 @@
...
+            is_little_endian = true;
+            iter:__builtin_get_next();
+            iter:__builtin_get_next();
+        }
+    }
+    return { iter, is_little_endian };
+}
+
+// --------- UTF8 -----------
+// https://www.unicode.org/versions/Unicode4.0.0/ch03.pdf#G31703
```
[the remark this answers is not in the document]
DaveDuck321: Whoops. I started with the 4.0 standard then realized it was out of date by the time I started implementing.

src/glang/lib/std/unicode.c3
```diff
@@ -0,0 +1,412 @@
...
+function enum_utf8_encoding_3_byte_E0 : () -> u32 = { return 0x03; }
+function enum_utf8_encoding_3_byte_E1_EC : () -> u32 = { return 0x04; }
+function enum_utf8_encoding_3_byte_ED : () -> u32 = { return 0x05; }
+function enum_utf8_encoding_3_byte_EE_EF : () -> u32 = { return 0x06; }
+function enum_utf8_encoding_4_byte_F0 : () -> u32 = { return 0x07; }
+function enum_utf8_encoding_4_byte_F1_F3 : () -> u32 = { return 0x08; }
+function enum_utf8_encoding_4_byte_F4 : () -> u32 = { return 0x09; }
+
+function get_utf8_encoding_length : (utf8_encoding_type: u32) -> isize = {
+    if utf8_encoding_type == enum_utf8_invalid() or utf8_encoding_type == enum_utf8_encoding_ascii() {
```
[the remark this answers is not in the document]
DaveDuck321: I've added a spec comment... It says to either die or replace with REPLACEMENT CHARACTER (which is what I do here)

src/glang/lib/std/unicode.c3
```diff
@@ -0,0 +1,412 @@
...
+
+    // U+0800..U+0FFF:      E0
+    if first_byte == 0xe0 {
+        return enum_utf8_encoding_3_byte_E0();
+    }
+
+    // U+1000..U+CFFF:      E1..EC
+    if is_in_range(first_byte, 0xe1, 0xec) {
+        return enum_utf8_encoding_3_byte_EE_EF();
+    }
```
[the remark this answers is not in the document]
DaveDuck321: Good spot. Interestingly these encodings were identical so this didn't change anything :upside_down_face:. Fixed

src/glang/lib/std/unicode.c3
```diff
@@ -0,0 +1,412 @@
...
+    if encoding_length == 1 {
+        return unsafe_decode_utf8_bytes({first_byte});
+    }
+
+    // Byte 2
+    if !self.underlying:has_next() {
+        return invalid_unicode_scalar(); // TODO: iterators should be nicer than this
+    }
+
+    let second_byte : u8 = self.underlying:peak();
```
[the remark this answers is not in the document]
DaveDuck321: wow, that was everywhere

src/glang/lib/std/unicode.c3
```diff
@@ -0,0 +1,412 @@
...
+    return unsafe_decode_utf8_bytes({first_byte, second_byte, third_byte, forth_byte});
+}
+
+@implicit has_next : (self : UTF8DecodeIter mut&) -> bool = {
+    return self.underlying:has_next();
+}
+
+function encode_to_utf8 : (scalar : UnicodeScalar, buffer: Span<u8>) -> View<u8> = {
+    if is_in_range(scalar.value, 0x00, 0x7f) {
+        buffer:get(0) = Narrow<u8>(scalar.value & 0x7f);
```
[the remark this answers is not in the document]
DaveDuck321: Yup, I like it tho :stuck_out_tongue_closed_eyes:

src/glang/lib/std/unicode.c3
DaveDuck321: UTF-16 has `2**16 + 2**20` codepoints (including the invalid surrogate pairs)... this needs 21 bits to represent in UTF-8 but anything past `0x10ffff` is invalid since it can't be represented in UTF-16.

src/glang/lib/std/unicode.c3
DaveDuck321: Yes. Good spot.

src/glang/lib/std/unicode.c3
```diff
@@ -0,0 +1,412 @@
...
+        return invalid_unicode_scalar();
+    }
+
+    if encoding_length == 1 {
+        return unsafe_decode_utf8_bytes({first_byte});
+    }
+
+    // Byte 2
+    if !self.underlying:has_next() {
+        return invalid_unicode_scalar(); // TODO: iterators should be nicer than this
```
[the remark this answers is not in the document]
DaveDuck321: Hmm, I think this comment was made out-of-date by the substitution algorithm. I still don't like `__builtin_get_next` everywhere but this line isn't awful.

src/glang/lib/std/unicode.c3
DaveDuck321: I think this should be a full API. Opened as an issue.

src/glang/lib/std/unicode.c3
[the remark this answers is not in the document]
antroseco: Reasonable

src/glang/lib/std/unicode.c3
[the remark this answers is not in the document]
antroseco: Sure

src/glang/lib/std/unicode.c3
antroseco: Ok, that makes sense

src/glang/lib/std/unicode.c3
```diff
@@ -354,6 +348,7 @@ function encode_to_utf8 : (scalar : UnicodeScalar, buffer: Span<u8>) -> View<u8>
 
     // Graphene cannot generate an invalid unicode scalar via `make<>` or the iterators
     // This is a bug in the user's code so I don't mind the unrecoverable error here
+    print(scalar.value);
```
antroseco, line 351: Leftover debug print!

antroseco (approved):

src/glang/lib/std/unicode.c3
DaveDuck321: Noooo!!!! But also `runtime_assert` should take a string to print :upside_down_face:

antroseco: One day!

tests/stdlib/unicode/unicode.c3
DaveDuck321: I've added a decoder fuzz test. This doesn't test for maximal subparts but does test the substitutions match the minimum criteria required by the standard.

DaveDuck321: I'm actually very happy with this test.... Its random but so far every subtle incorrectness I've (intentionally) introduced has been caught on every run.

DaveDuck321: I could probably make it smaller without much risk... It only takes 100ms compiled on my machine tho

PR #171, opened 2024-10-15
Fix the CI lint stage

DaveDuck321:
The `ubuntu-latest` image update to a version that prevent `pip --user` (and tells us to use `apt`). We can work around this by creating virtual environments.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -32,11 +32,11 @@ jobs:
         run: |
           python3 -m venv "${HOME}/venv"
           source "${HOME}/venv/bin/activate"
-          python3 -m pip install hatch==1.9.4
+          pip install hatch==1.9.4
         shell: bash
       - name: Run ruff
         run: |
-          source "${HOME}/venv/bin/activate
+          source "${HOME}/venv/bin/activate"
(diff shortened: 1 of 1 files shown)


DaveDuck321:
@antroseco Actually looking at #170 I see that you've already fixed this using the `setup-python` action.

I don't want to leave CI red so I'll merge this rn but feel free to revert with your changes if there's any reason to prefer yours.

antroseco:
Oh, I'm guessing it's the ruff step that failed. I have a slight preference for `setup-python` cause that let's you control the version too, instead of relying on system python.

DaveDuck321:
> This was already fixed upstream - I'm surprised we're still seeing it: [isort/isort-action@`v1.1.1`/CHANGELOG.md#111---2024-10-14](https://github.com/isort/isort-action/blob/v1.1.1/CHANGELOG.md?rgh-link-date=2024-10-15T22%3A11%3A00Z#111---2024-10-14)

Yeah, I'd have hoped it was pulled in automatically but both were broken here: https://github.com/DaveDuck321/GrapheneLang/commit/67df2217fb5e53eb4911b77023f94f18b0cfd0ff. Manually updating the version number seemed to do the trick tho.

PR #172, opened 2024-10-18
Add various `builtin`s for embedded programming

DaveDuck321:
This PR adds the following builtins:
- `__builtin_trap`
- `__builtin_unreachable`  (which both allows both graphene and llvm to omit function returns)
- `__builtin_volatile_write`
- `__builtin_volatile_read`

Thoughts:
- Maybe `unreachable` can be a first-class graphene statement? Atm if we tried wrapping it in a standard library function we would still get "control flow reaches the end of function" errors.
- A rich volatile that has an allowed reordering hierarchy would be cool but LLVM doesn't support this:
   - eg. ofc this specific syntax is shit and would need some thinking
   - `volatile_write<.0>()` and `volatile_write<.0>()` couldn't be reordered
   - `volatile_write<.0>()` and `volatile_write<.1>()` could be reordered
   - `volatile_write<.0.0>()` and `volatile_write<.1.0>()` could be reordered
   - `volatile_write<.0>()` and `volatile_write<.0.0>()` couldn't be reordered
   -  `volatile_write<.0.0>()` and `volatile_write<.0.0.0>()` couldn't be reordered
   -  `volatile_write<.0.0.1>()` and `volatile_write<.0.0.0>()` could be reordered


src/glang/codegen/builtin_callables.py
```diff
@@ -698,6 +700,198 @@ def assert_can_write_to(self) -> None:
...
+            meta,
+        )
+
+    def __repr__(self) -> str:
+        return "TrapExpression()"
+
+    def generate_ir(self, ctx: IRContext) -> IROutput:
+        ir = IROutput()
+        dbg = self.add_di_location(ctx, ir)
+        ir.lines.append(f"call void @llvm.trap(), {dbg}")
```
antroseco, line 725: What does this lower to on x86? From the reference
> This intrinsic is lowered to the target dependent trap instruction. If the target does not have a trap instruction, this intrinsic will be lowered to a call of the abort() function.

Noting that we don't have an `abort()` function

src/glang/codegen/builtin_callables.py
```diff
@@ -698,6 +700,198 @@ def assert_can_write_to(self) -> None:
...
+        )
+
+    def __repr__(self) -> str:
+        return f"VolatileWriteExpression({self._to_write} to {self._addr})"
+
+    def generate_ir(self, ctx: IRContext) -> IROutput:
+        # Remove any indirect references
+        conv_to_write, extra_exprs_to_write = do_implicit_conversion(
+            self._to_write, self._to_write.result_type
+        )
```
antroseco, line 810: Did we not have a helper for that?

src/glang/codegen/builtin_callables.py
```diff
@@ -698,6 +700,198 @@ def assert_can_write_to(self) -> None:
...
+        specialization: list[SpecializationItem],
+        arguments: list[TypedExpression],
+        meta: Meta,
+    ) -> None:
+        assert len(specialization) == 0
+        assert len(arguments) == 1
+
+        self.result_ref = None
+
+        (self._addr,) = arguments
```
antroseco, line 852: This is kinda cursed lol

src/glang/codegen/builtin_callables.py
```diff
@@ -698,6 +700,198 @@ def assert_can_write_to(self) -> None:
...
+        ir = self.expand_ir(extra_exprs_addr, ctx)
+
+        self.result_ref = f"%{ctx.next_reg()}"
+
+        dbg = self.add_di_location(ctx, ir)
+
+        # <result> = load volatile <ty> <value>, ptr <pointer>
+        ir.lines.append(
+            f"{self.result_ref} = load volatile {self.result_type.ir_type}, "
+            f"{conv_addr.ir_ref_with_type_annotation}, {dbg}",
```
antroseco, line 879:
```suggestion
        # <result> = load volatile <ty>, ptr <pointer>
        ir.lines.append(
            f"{self.result_ref} = load volatile {self.result_type.ir_type}, "
            f"{conv_addr.ir_ref_with_type_annotation}, {dbg}",
```
Something doesn't look right here

src/glang/codegen/builtin_callables.py
```diff
@@ -698,6 +700,198 @@ def assert_can_write_to(self) -> None:
...
+    @property
+    def ir_ref_without_type_annotation(self) -> str:
+        assert self.result_ref is not None
+        return self.result_ref
+
+    def assert_can_read_from(self) -> None:
+        pass
+
+    def assert_can_write_to(self) -> None:
+        raise OperandError("Cannot assign to `__builtin_volatile_write()`")
```
antroseco, line 892:
```suggestion
        raise OperandError("Cannot assign to `__builtin_volatile_read()`")
```

src/glang/lib/std/aarch64_linux/syscalls.c3
```diff
@@ -68,3 +68,8 @@ function sys_munmap : (addr: iptr, length: isize) -> i64 = {
 function sys_exit : (code : int) -> void = {
     _syscall_1(code, /* exit */ 93);
 }
+
+function sys_assert_failed: () -> void = {
```
antroseco, line 72: Does this really need to be a target-specific function?

src/glang/lib/std/volatile.c3
```diff
@@ -0,0 +1,71 @@
...
+
+function [T] write : (var: Volatile<T>&, to_write_data: T) -> void = {
+    volatile_write(&mut var.underlying, to_write_data);
+}
+
+function [T] make<Volatile<T>> : (data: T mut&) -> Volatile<T> = {
+    return {&mut data};
+}
+
+typedef [T] ReadVolatile : {
```
antroseco, line 17: What's the use case for this and the other two structs below?

antroseco (approved):

antroseco:
> Thoughts:
> 
>     * Maybe `unreachable` can be a first-class graphene statement? Atm if we tried wrapping it in a standard library function we would still get "control flow reaches the end of function" errors.

I wouldn't be against it, but let's see how it goes

>     * A rich volatile that has an allowed reordering hierarchy would be cool but LLVM doesn't support this:
>       
>       * eg. ofc this specific syntax is shit and would need some thinking
>       * `volatile_write<.0>()` and `volatile_write<.0>()` couldn't be reordered
>       * `volatile_write<.0>()` and `volatile_write<.1>()` could be reordered
>       * `volatile_write<.0.0>()` and `volatile_write<.1.0>()` could be reordered
>       * `volatile_write<.0>()` and `volatile_write<.0.0>()` couldn't be reordered
>       * `volatile_write<.0.0>()` and `volatile_write<.0.0.0>()` couldn't be reordered
>       * `volatile_write<.0.0.1>()` and `volatile_write<.0.0.0>()` could be reordered

I'm not sure I understand the syntax here. What does `<.1.0`> mean?

src/glang/codegen/builtin_callables.py
DaveDuck321: https://godbolt.org/z/E14s5zK9v which gives a https://mudongliang.github.io/x86/html/file_module_x86_id_318.html. Which triggers a `SIGILL (Illegal instruction)` on linux

Looks like arm also has one: https://godbolt.org/z/

src/glang/codegen/builtin_callables.py
DaveDuck321: I'm not sure. I have a fuzzy memory of one... but also I can't find it... maybe this is already the helper?

src/glang/lib/std/aarch64_linux/syscalls.c3
DaveDuck321: I did this because I still want asserts on the gameboy but also I liked your changes from #170 better than just trapping everywhere. Moving this into the target dir seemed like a decent compromise.

I think in the long term we should be able to list target include directories in the actual target definition so all linux targets could do something like `include: posix` and we wouldn't need to duplicate this utility.

DaveDuck321:
> I'm not sure I understand the syntax here. What does `<.1.0`> mean?

It was meant to be like a tree so something like:

`volatile_write<global.component.sub_component>`

All changes to the global state would be volatile with respect to everything.
Changes to component state would only be volatile with respect to members of the same component.
Changes to the sub component would only be volatile in this specific sub-component.

The logic would be something like:
- Turning on the address bus (global) will effect the whole system
- Turning on the graphics unit (the component) will only effect graphics. It should be done before any other graphics writes and after the address bus is enabled... but ordering wrt the IO controller (another subcomponent) doesn't matter.

src/glang/lib/std/volatile.c3
DaveDuck321: For write volatile, I think something like this makes sense:
```graphene
// Memory mapped on the GPU... but also the GPU will never change this value
WriteVolatile<int> screen_scroll = { MAGIC_ADDRESS };

// Tell the GPU to update the scroll... This write MUST happen
screen_scroll:write(20);

// We need to do some math with it (if its already in a register we might as well save time and use that)
let player_position_on_screen : int =  player_position - screen_scroll:read();
```

I'm not sure I can argue for `ReadVolatile` as well.... If it supported member access of large structs you could make an argument someone might DMA into a struct (that you also want to update in-place for some reason). Maybe `ReadOnlyVolatile` makes more sense... it wouldn't have a `write` member.

I added `NonVolatile` just so SFINAE that normally takes volatile types can be written to accept non-volatile types. I'll delete this if I end up changing `ReadVolatile` to `ReadOnlyVolatile`.

src/glang/codegen/builtin_callables.py
DaveDuck321: Black ruins it. I think this is cooler as the `,=` operator
```
self._addr ,= arguments
```

src/glang/codegen/builtin_callables.py
antroseco: What about `dereference_double_indirection()`?

src/glang/lib/std/aarch64_linux/syscalls.c3
antroseco: Ok that's fair enough

src/glang/codegen/builtin_callables.py
DaveDuck321: That seems like a misuse of the utility since we only have (at most) a single layer of indirection here.

This would manifest if we tried something like `__builtin_volatile_write(&addr, 6)` you don't want `dereference_double_indirection` to try and deference `6`. I think they'd produce identical IR for anything else tho.

src/glang/lib/std/volatile.c3
DaveDuck321: I've removed `ReadVolatile` and `NonVolatile`

antroseco:
> > I'm not sure I understand the syntax here. What does `<.1.0`> mean?
> 
> It was meant to be like a tree so something like:
> 
> `volatile_write<global.component.sub_component>`
> 
> All changes to the global state would be volatile with respect to everything. Changes to component state would only be volatile with respect to members of the same component. Changes to the sub component would only be volatile in this specific sub-component.
> 
> The logic would be something like:
> 
>     * Turning on the address bus (global) will affect the whole system
> 
>     * Turning on the graphics unit (the component) will only affect graphics. It should be done before any other graphics writes and after the address bus is enabled... but ordering wrt the IO controller (another subcomponent) doesn't matter.

Ah, I see. I assume this is something that will prevent compiler reordering, but we would still need to add memory barriers (implicitly?) if you want to prevent CPU reordering too.

Do you foresee any performance improvements from this finer-grained control?

src/glang/codegen/builtin_callables.py
antroseco: Yeah, you're right

src/glang/codegen/builtin_callables.py
antroseco: lmao, that's a new one

PR #176, opened 2024-11-30
Improve the test runner and switch to `hatch test`

antroseco:
The test runner now accepts a list of paths, and like `pytest`, these can be either a test file or directory to search for tests. If no paths are provided, then it searches the `tests/` directory instead (the existing behaviour).

I've also (finally) fixed that bug where the test config parser would crash if the file was not a test.

This and some project changes let us use `hatch test [args...]` to launch the test runner, instead of `hatch run dev:test [args...]` that we have to do today.

### DIFF
--- a/.github/workflows/graphene_tests.yml
+++ b/.github/workflows/graphene_tests.yml
@@ -10,7 +10,7 @@ jobs:
     strategy:
       matrix:
         llvm-version: [17, 18]
-        python-version: [3.12]
+        python-version: [3.13]
         runner: [ubuntu-latest, buildjet-2vcpu-ubuntu-2204-arm]
 
     runs-on: ${{ matrix.runner }}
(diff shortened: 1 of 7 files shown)


tests/run_tests.py
```diff
@@ -205,14 +206,23 @@ def run_test_throw_on_fail(config: TestConfig, file_path: Path) -> None:
...
+    # Try and shorten the path if possible.
+    test_name = (
+        str(test_path.relative_to(PARENT_DIR))
+        if test_path.is_relative_to(PARENT_DIR)
+        else str(test_path)
+    )
+
+    config = parse_file(test_path)
+    if config is None:
+        print(f"SKIPPED '{test_name}' (not a test)")
```
DaveDuck321, line 219: This might get interleaved. Better take the `io_lock`

tests/run_tests.py
```diff
@@ -303,32 +316,34 @@ def build_jit_dependencies() -> None:
...
+            all_test_files.update(
+                test_file.resolve(strict=True)
+                for test_file in file_or_dir.rglob("**/*.c3")
+                if not test_file.stem.startswith("_") and test_file.is_file()
+            )
+        else:
+            print(f"ERROR: Invalid path {file_or_dir}")
 
-        sys_exit(run_tests(all_test_files, args.workers))
+    sys_exit(run_tests(sorted(all_test_files), args.workers))
```
DaveDuck321, line 346: While this is objectively an improvement I did like it when they were sorted chronologically and my new test was at the bottom.

tests/test_config_parser.py
```diff
@@ -121,12 +121,16 @@ def grep_ir_cmd(self, token: Token) -> None:
...
-def parse_file(path: Path) -> TestConfig:
+def parse_file(path: Path) -> TestConfig | None:
     with path.open(encoding="utf-8") as file:
         lines = [
             line.removeprefix("///").strip() for line in file if line.startswith("///")
         ]
 
+    if not lines:
+        # This is not a test file.
+        return None
```
DaveDuck321, line 132: Hmm, this will still crash if the syntax of the testing comments is wrong right?

This is still better than before tho

.github/workflows/graphene_tests.yml
```diff
@@ -25,22 +25,22 @@ jobs:
...
-      run: |
-        wget https://apt.llvm.org/llvm.sh
-        chmod +x llvm.sh
-        sudo ./llvm.sh ${{ matrix.llvm-version }}
-        sudo apt-get install -y llvm-${{ matrix.llvm-version }}-runtime
-    - name: Fetch tags
-      run: git fetch --tags --force origin
-    - name: Run tests
-      run: hatch run dev:test
+      - name: Checkout
```
DaveDuck321, line 28: Any reason for the indentation changes? I'm not really sure whats considered good yml tbh

DaveDuck321 (approved):
:star_struck:

tests/run_tests.py
antroseco: Woops, good spot

tests/run_tests.py
antroseco: I'm not sure it's worth replicating that and I don't want tests to run in a non-deterministic order :sweat_smile:

tests/test_config_parser.py
antroseco: Yes, which is the right thing to do I believe (we're not testing the testing commands). I've improved this now to print the file path and the message, but not the whole stack trace.

.github/workflows/graphene_tests.yml
antroseco: Ugh no, just vscode decided that this looks better now. Does yours prefer the old indentation?

PR #177, opened 2024-12-18
Move to llvm-19 and debug records

antroseco:
No changes are strictly necessary to move to llvm-19, although debug records are now the default.
(see https://llvm.org/docs/RemoveDIsDebugInfo.html)

I decided to just switch to the new textual IR format. The changes are trivial but they break llvm-17 and llvm-18 support. Up to you whether this is okay or we want to hold off on this change.


DaveDuck321 (commented):
Wow this was such a small change! Manjaro's still on llvm 18 so lets hold off for a bit tho

DaveDuck321:
Manjaro has llvm 19. Let's do this!

antroseco:
It's time for LLVM 20 already!

PR #179, opened 2024-12-26
Driver: allow argument passthrough

DaveDuck321:
Changes:
1) `--extra-opt-args` and `--extra-llc-args` allow us to pass arguments directly to a specific stage (eg. `--strip-debug` or target-specific options)
2) Add support for semantic optimization levels like `-Oz` or `-Og`
3) Always print `stderr` from the various tools (useful for debugging, they shouldn't print anything on success anyway)
4) Expand the assembly macros with a target.... This is an annoying bug with clang but if you don't give a target, it'll run for under the host target, if clang doesn't support the host it'll error even when expanding macros.
5) Add a "verbose" driver mode


src/glang/driver.py
```diff
@@ -31,15 +31,32 @@ def run_checked(args: list[str | Path], stdin: bytes | None = None) -> bytes:
         capture_output=True,
         check=False,
     )
-    if result.returncode != 0:
+    if result.stderr:
         command_str = " ".join(str(arg) for arg in args)
```
antroseco, line 35: You didn't touch this lines, but this looks like a job for `shlex.join` instead of `str.join`

src/glang/driver.py
```diff
@@ -214,6 +240,7 @@ def assemble(self, args: DriverArguments) -> Assembly:
                 f"--mtriple={get_target_triple()}",
                 "--filetype=asm",
                 f"-O{args.optimize}",
+                *(args.extra_llc_args.split(",") if args.extra_llc_args else []),
```
antroseco, line 243: Up to you, but you could set a custom `type` (https://docs.python.org/3/library/argparse.html#type) for these two arguments that splits them for you (and then set the default to an empty list)

antroseco (approved):

